Issue #42, opened 2018-05-28
Proof of concept: Recreate AsimovOperation GUI using flask


PeterGhimself:
## some useful links
docs: http://flask.pocoo.org/
bootstrap: https://colorlib.com/wp/free-bootstrap-admin-dashboard-templates/
flask admin dashboard: https://github.com/jonalxh/Flask-Admin-Dashboard.git
--> `git clone https://github.com/jonalxh/Flask-Admin-Dashboard.git`

bnduwi:
The GUI flask POC is mainly done. Right now it looks like this. What is left is to add buttons or styling changes for when manual switch is turned on and off. 

For @gabrielkarras here is the screen shot needed to see the manual controls:
![image](https://user-images.githubusercontent.com/32171245/43234228-df211bc6-9048-11e8-9b3e-eab559ea26e1.png)

Todo:
- change layout of buttons when switched to manual control = off. Use javascript to play around with css to make layouts hidden/visible.
- add a placeholder for the arm vision.

Current issues:
- When resized the containers in the bottom row don't become responsive or don't stack on top of each other


Currently the GUI looks like this:
![image](https://user-images.githubusercontent.com/32171245/43234375-aa02ca24-9049-11e8-9848-932a2853a4c6.png)

PR #49, opened 2018-07-13
Virtualenv setuptools imports 48

### LINKED ISSUE #48: Setup virtualenv, setuptools, and proper pytest directory structure
This issue is going to be somewhat loaded but it can't really be fragmented due to the heavy coupling between setting up the `pytest` structure and `virtualenv`.

Two essential aspects are:
- We need to ensure we are using virtual environments (`virtualenv`) and install only the packages the project needs as specified via a `requirements.txt` file (which is used by `pip` to install them all)
- For `pytest` to work properly, we need to alter the structure of the project a bit and setup `setuptools` which will require a `setup.py` file

tatumalenko:
closes #48 

All setup, have a gander and pull puh-lease so I can move forward with the linting/formatting :)


tatumalenko:
Ugh.. Seems Travis can't install `pygobject` without having the necessary system dependencies installed first. Is there a way to make Travis install a package via its own OS wide virtual package manager as instructed [here](http://pygobject.readthedocs.io/en/latest/getting_started.html)?

README.md
```diff
@@ -1,19 +1,121 @@
 # robotics-prototype
-This repo contains the beginning of the new (2017-2018) robotics software team code
+This repo contains the beginning of the new (2017-2018) Robotics software team code.
 
-## How to upload arduino scripts from the odroid
+## Contributing and Development Environment Instructions
+Firstly, this project is built in Python 3.3+. You need to have a version of Python installed that is 3.3+. Make sure that whenever you use `python` or `python3` or whatever later on meets this requirement.
```
PeterGhimself: GTK code is in 2.7, even though we are moving away from that and python 2 in general

PeterGhimself:
@tatumalenko I'm not sure right off the bat, I'd have to look into it. But since we are moving away from GTK is it worth the trouble?

README.md
tatumalenko: @PeterGhimself I was able to run everything using Gtk3 and Python 3+. I guess the question is should we bother putting it in the requirements.txt? I looked into setting up Travis with `apt-get install <package>` directives, it seems possible, might be more trouble than its worth, though.

PeterGhimself: I mean, if it's not too much trouble to figure out (like if those `apt-get install` directives work out), then it's fine for now I guess, since #42 is still under development and not merged in with master yet. If's too much of a hastle to figure out I guess we can omit it from `requirements.txt`.

tatumalenko: @PeterGhimself Ok so I looked into this ad nauseam (trying to get `PyGObject` and `python3-gi` to work together with Travis-CI).

My conclusion is the latest version of `PyGObject` (3.28.3) requires a version of `python3-gi` too recent for the latest one available on Trusty. This is seen in the screenshot where `gobject-introspection-1.0 >= 1.46.0` is requested by `PyGObject`, but Trusty (i.e. Ubuntu 14.04) only supports up to 1.40.0 as depicted [here](https://launchpad.net/ubuntu/+source/gobject-introspection). In other words, TravisCI would need to offer a Ubuntu version at minimum Xenial (16.04+) in which 1.46.0 would be available (but it doesn't).

image

More reason to move on to Flask!!

PeterGhimself: 💯 👌 🔥

PR #50, opened 2018-07-27
Setup pylint yapf 47

### LINKED ISSUE #47: Setup pylint & yapf
Getting a basic [`pylint`](https://github.com/PyCQA/pylint) & [`yapf`](https://github.com/google/yapf) setup going so that everyone can adopt the same coding style guidelines.

The basic additions that will be incorporated are:
- `.pylintrc` file will live in the root directory and store all linting configs (currently same as Google's)
- `.vscode` folder will store workspace settings for those of us using **VSCode** that we should all have in common (please don't put your pc specific settings in there such as `pythonPath`, etc)
   - This folder will essentially contain directives that tell **VSCode** which linter and formatter to use (if either one of them isn't installed, **VSCode** will install it for you)

For those using **PyCharm**, I'm not too sure of the setup involved but according to [this](https://stackoverflow.com/questions/38134086/how-to-run-pylint-with-pycharm) it shouldn't be too painful to get the linter (`pylint`) running. I haven't looked into how **PyCharm** can use the formatter (`yapf`) yet.

**NOTE:** Although the use of Python Docstrings as a rule won't be set in the linters settings, I really want to emphasize that instead of using the `#` for comments, we should be using triple double quotes `"""This is a one-liner docstring"""` (i.e. Docstrings). [Here](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings) is a reference for conventions used a **Google** around good comments/Docstrings.

tatumalenko:
Linting and formatting using `pylint` and `yapf` are set up and ready to go. Minor additions to be honest. Have a look at the `Formatting Guide` section under `Contributing and Development Environment Instructions` in the `README` file for an overview.

Closes #47


bnduwi:
Test on Windows with Peter solved many issues with Pylint and yapf. It all works now on window.

PR #64, opened 2018-10-07
Odroid stress test 62

### LINKED ISSUE #62: Odroid Stress Test
Simple script to test the odroid's performance while CPU is overclocking for variable amount of time.

Start by reading [this](https://core-electronics.com.au/tutorials/stress-testing-your-raspberry-pi.html) for more info.
Further info on [stress](https://www.hecticgeek.com/2012/11/stress-test-your-ubuntu-computer-with-stress/) debian package.

TODO:
- [x] Get CPU temp/clock freq
- [x] Write script to run for 60s or specified time and print temp/freq every second
- [x] Port bash script(s) to python
- [x] Add final report: print min/max/avg statistics when stress test finishes
- [x] Add file logging
- [x] Test confirming everything work on odroid C1+, XU4, XU4Q
- [x] Refactor, fix minor bugs, leave acknowledgment for "known issues" that won't be handled
- [x] Final test re-confirming everything works on odroid (since developing/testing on desktop ubuntu for convenience first)
- [x] Update docs (README), mention current/power draw would be best measured with physical ammeter rather than purely through software

PeterGhimself:
closes #62 

## Summary
- The script `odroid_stresser.py` is able to be run for a variable amount of seconds (default 60), reports the temperature and clock frequency of the CPU to the user every second, and logs it to a file (creating a new one on each run).
- Updated some stuff for pylint, yapf
- Travis will now only run tests in tests/unit folder
- Updated some READMEs for my demos
- Added some stuff to gitignore

Check the README for more details regarding the `odroid_stresser` script  :sunglasses:


robot/util/stresstest/.bash_cmds.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
tatumalenko: I'm just thinking out loud here, but why is this `stresstest` inside the `util` section of the project?

tatumalenko: I see! I sort of find it would be feel more appropriate to be within the `tests/benchmarks` or something of the sort as seen [here](https://github.com/kriasoft/Folder-Structure-Conventions), but its a minute detail tbh.

robot/util/stresstest/README.md
```diff
@@ -1,8 +1,50 @@
...
+Max CPU temp: 52.000 C
+Avg CPU temp: 48.6 C
+
+Clock speed stats:
+Min CPU freq: 2451.913 MHz
+Max CPU freq: 3192.718 MHz
+Avg CPU freq: 3055.7672 MHz
+```
+
+## Bash Setup (optional)
```
tatumalenko: I'm wondering at this point, if the test is slightly outdated (and will likely be attributed no attention moving forward if additional things change), is there a point to keep it? I feel it would almost be safer to discard in case some functionality is likely to miss and considering we're using Python almost in its entirety for this project, we might as well stay true to it and keep the focus? Again, a minute detail..

PeterGhimself: fair point
My logic for leaving it behind was because it might be convenient to use the bash functions from the console for quick testing of them, if you set up to source the bash functions in your `.bashrc` or just source the `bash_cmds.sh` if you just want to temporarily have the commands at your disposal

robot/util/stresstest/.bash_cmds.sh
PeterGhimself: Moved it to new folder: `/tests/benchmarks/`, thanks!

PR #69, opened 2018-11-22
Skid steering demo 65

### LINKED ISSUE #65: Skid steering demo
Using arduino, 6 servos (blue boys), wireless commands (bluetooth), implementing proof of concept for skid steering system.
Later to be modified for DC motors and ROS.

fauziame:
closes #65

### DIFF
--- a/robot/demos/SkidSteerServos/README.md
+++ b/robot/demos/SkidSteerServos/README.md
@@ -1,0 +1,9 @@
+# Skid Steering Demo
+
+## Setup
+
+You will need to import the library ArduinoBlue for bluetooth connection with android/iOS. You may use the Arduino IDE library manager: 
+
+- Navigate to: Sketch > Include Library > Manage Libraries...
+- Search for 'arduinoblue'
+- Install
--- a/robot/demos/SkidSteerServos/SkidSteerServos.ino
+++ b/robot/demos/SkidSteerServos/SkidSteerServos.ino
@@ -11,17 +11,11 @@ Servo myservo3;
 Servo myservo4;
 Servo myservo5;
 
-int xPin = A1;
-int yPin = A0;
-int buttonPin = 2;
-
-int xPosition = 0;
-int yPosition = 0;
-int buttonState = 0;
 
 int prevThrottle = 49;
 int prevSteering = 49;
 int throttle, steering;
+// Initialize Left & Right Velocity variables to Rest
 int vr = 90;
 int vl = 90;
 SoftwareSerial bluetooth(7, 8);
@@ -29,10 +23,13 @@ SoftwareSerial bluetooth(7, 8);
 ArduinoBlue phone(bluetooth);
 
 void setup() {
-  // put your setup code here, to run once:
-  myservo.attach(9);  // attaches the servo on pin 9 to the servo object
-  myservo1.attach(10);  // attaches the servo on pin 9 to the servo object
+  // Attach Servos to pins:
+  // Right side servos
+  myservo.attach(9);  
+  myservo1.attach(10);  
   myservo2.attach(11);
+  
+  //Left side Servos
   myservo3.attach(3);
   myservo4.attach(5);
   myservo5.attach(6);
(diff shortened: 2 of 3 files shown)


robot/demos/SkidSteerServos/SkidSteerServos.ino
```diff
@@ -0,0 +1,107 @@
+#include <ArduinoBlue.h>
```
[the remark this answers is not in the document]
vashmata: If anything you could have some kind of grouping. For example, grouping all the communications header files together, then grouping all the motor control headers together, but separating these groups with a newline.

robot/demos/SkidSteerServos/README.md
```diff
@@ -0,0 +1,9 @@
+# Skid Steering Demo
+
+## Setup
```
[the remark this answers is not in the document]
vashmata: I remember seeing a more detailed description somewhere listing all of the equipment used. That would be good to put here, like Peter said.

robot/demos/SkidSteerServos/SkidSteerServos.ino
```diff
@@ -0,0 +1,107 @@
...
+#include <SoftwareSerial.h>
+
+#include <Servo.h>
+
+Servo myservo;
+Servo myservo1;
+Servo myservo2;
+Servo myservo3;
+Servo myservo4;
+Servo myservo5;
```
vashmata:
```suggestion
Servo servoFrontRight;
Servo servoMiddleRight;
Servo servoBackRight;
Servo servoFrontLeft;
Servo servoMiddleLeft;
Servo servo BackLeft;
```

robot/demos/SkidSteerServos/SkidSteerServos.ino
```diff
@@ -0,0 +1,107 @@
+#include <ArduinoBlue.h>
+
+#include <SoftwareSerial.h>
+
+#include <Servo.h>
+
+Servo myservo;
```
vashmata: Like Peter said, it's good to have descriptive variable names for when the code gets more complex
```suggestion
Servo servoFrontRight;
Servo servoMiddleRight;
Servo servoBackRight;
Servo servoFrontLeft;
Servo servoMiddleLeft;
Servo servo BackLeft;
```

robot/demos/SkidSteerServos/SkidSteerServos.ino
```diff
@@ -0,0 +1,107 @@
...
+ // New Rest Value is 0
+  throttle -= 49;
+  steering -= 49;
+     
+  delay(200); // add some delay between reads
+
+// If statement for CASE 1: steering toward the RIGHT 
+  if (steering <= 0 ) {
+    deg = mapFloat(steering, -49, 0, -1, 1);
+    vr = map(throttle*deg, -49, 49,0,180);
```
vashmata: It might be possible to use floats from 0.0 to 180.0 for Servo objects, which may give better accuracy for PWM control. Personally I don't even bother with using Servo objects when I'm controlling continuous servos because it kind of defeats the purpose when the 180 doesn't even correspond to an angle anymore.

robot/demos/SkidSteerServos/SkidSteerServos.ino
```diff
@@ -0,0 +1,107 @@
...
+    deg = mapFloat(steering, 0, 49, 1, -1);
+    vr = map(throttle, -49, 49, 0,180);
+    vl = map((-1)*throttle*deg, -49, 49,  0,180);
+
+} 
+  
+// Print Velocity Values   
+  if (prevThrottle != throttle || prevSteering != steering) {
+    Serial.print("Throttle: "); Serial.print(throttle); Serial.print("\tSteering: "); Serial.println(steering);
+    Serial.print(steering);Serial.print('>'); Serial.print(throttle);Serial.print('o');
```
vashmata: In the grand scheme of things it probably doesn't make a difference whether you print a char or a String so you might as well go with a String to keep things more uniform... Doesn't really make a difference here but sometimes I get errors when I mix up chars with Strings.
```suggestion
    Serial.print(steering);Serial.print(">"); Serial.print(throttle);Serial.print("o");
```

robot/demos/SkidSteerServos/SkidSteerServos.ino
```diff
@@ -0,0 +1,107 @@
...
+ // Lead Velocity Value from bluetooth controller. Values range from 0 to 90
+  throttle = phone.getThrottle();
+ // Steering Value from bluetooth controller. Values range from 0 to 90
+  steering = phone.getSteering();
+ 
+ // New Rest Value is 0
+  throttle -= 49;
+  steering -= 49;
+     
+  delay(200); // add some delay between reads
```
vashmata: Just a heads up that when you move over to the real project, we'll be using encoders which will be triggered by interrupt service routines, so we won't be able to use delay() anymore. The classic way of getting around this is by using millis() or micros()... Teensy makes things even easier with the elapsedMillis (possibly elapsedMicros too?) construct.

robot/demos/SkidSteerServos/SkidSteeringDC.ino
```diff
@@ -0,0 +1,93 @@
+#include <ArduinoBlue.h>
+#include <SoftwareSerial.h>
+#include <Servo.h>
+
+int prevThrottle = 49;
+int prevSteering = 49;
+int throttle, steering;
+// Initialize Left & Right Velocity variables to Rest
+int VR = 127;
```
vashmata: You'll see in my code that I actually make a macro definition for the stopping speed because I've seen that in some cases (mostly with continuous servos that you can tune manually) it's not necessarily a value of 127 that stops the motor. Makes it nicer to port the code to new things if you can change the value at the top and it fixes it everywhere in your code.

robot/demos/SkidSteerServos/SkidSteeringDC.ino
```diff
@@ -0,0 +1,93 @@
...
+  Serial.begin(9600); 
+  bluetooth.begin(9600);
+
+  //activate pull-up resistor on the push-button pin
+  pinMode(buttonPin, INPUT_PULLUP); 
+  
+  Serial.println("setup complete");
+}
+
+float deg = 0;
```
vashmata: I could have mentioned it for the other file as well but I find it a little weird that you make new variable definitions between setup() and loop(). I'd put everything at the top.

robot/demos/SkidSteerServos/SkidSteeringDC.ino
```diff
@@ -0,0 +1,93 @@
...
+
+
+// New map function that supports float variables.
+ float mapFloat(float x, float in_min, float in_max, float out_min, float out_max) {
```
vashmata: Similar to what I said about the variable definition, it's nicer to either define your functions right before setup()... Or, what I and lots of others like to do, is to simply declare the function right before setup() and then actually define it after loop(). This hides away the details until the person trying to read your code feels the need to delve deeper and understand what's really going on with the function.

The forward declaration is standard C practice and it seems like it's not always necessary in Arduino to declare it before if you define it later, but I like to do so anyway.

robot/demos/SkidSteerServos/SkidSteeringDC.ino
```diff
@@ -0,0 +1,93 @@
...
+int VR = 127;
+int VL = 127;
+SoftwareSerial bluetooth(7, 8);
+
+ArduinoBlue phone(bluetooth);
+
+void setup() {
+  // Attach Servos to pins:
+  // Right side servos
+  DC_PIN_1 = 9;
```
vashmata:
```suggestion
  DC_PIN_FRONT_RIGHT = 9;
  DC_PIN_MIDDLE_RIGHT = 10;
  DC_PIN_BACK_RIGHT = 11;
  
  // Left side DC motors
  DC_PIN_FRONT_LEFT = 3;
  DC_PIN_MIDDLE_LEFT = 5;
  DC_PIN_BACK_LEFT = 6;
```

vashmata: Just a suggestion. If you hate my nomenclature you can pick your own or whatever is used in the litterature for 6-wheeled vehicles

robot/demos/SkidSteerServos/SkidSteerServos.ino
PeterGhimself: How about `_delay_ms()` from including `<util/delay.h>`? :eyes:

robot/demos/SkidSteerServos/SkidSteeringDC.ino
PeterGhimself: I think when we define our own standards for C/C++/Arduino we should use forward declarations and define the functions at the bottom always.

robot/demos/SkidSteerServos/SkidSteerServos.ino
vashmata: doesn't that do the same thing? I think the only way around it is to do the classic:
```
  unsigned int prevRead = millis();
  if(millis()-prevRead > 200){
    // do thing
  }
```
which can be written as the following with the proper library (or on a Teensy it comes built in):
```
  elapsedMillis prevRead;
  if(prevRead > 200){
    // do thing
    prevRead = 0;
  }
```

robot/demos/SkidSteerServos/SkidSteerServos.ino
```diff
@@ -0,0 +1,107 @@
...
+
+// If statement for CASE 1: steering toward the RIGHT 
+  if (steering <= 0 ) {
+    deg = mapFloat(steering, -49, 0, -1, 1);
+    vr = map(throttle*deg, -49, 49,0,180);
+    vl = map((-1)*throttle, -49, 49,  0,180);
+  }
+
+// If statement for CASE 1: steering toward the LEFT 
+  if (steering >= 0 ) {
```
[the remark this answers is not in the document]
fauziame: Its not necessary to have a forward steering case because if steering = 0 then our deg will be 1 which wouldn't really make a change in velocities between the wheels.

For this case, I removed one of the =

robot/demos/SkidSteerServos/SkidSteerServos.ino
fauziame: Yea I think that's reasonable, I just wasn't sure to what degree the motors are sensitive to these values. However, I updated it to use floats.

robot/demos/SkidSteerServos/SkidSteeringDC.ino
fauziame: Lol yea I see what you mean, I think its just a bad habit. Fixed!

PR #70, opened 2018-12-01
View camera feed 22

### LINKED ISSUE #22: Users shall be able to view the camera feed
Use cases:
- 4.1.11 Enable camera feed
This is a button to allow the user to enable video stream. When enabled this also includes the action of allowing the camera to send the feed to the GUI application.
- 4.1.12 Disable camera stream
This is a button to allow the user to disable video stream. This is useful since it allows the number of requests and data received to be reduced, which could improve performance when needed.
- 4.1.15 Update camera FPS
The camera feed has the potential of causing performance issues due to the number of data being fed to the GUI application. To reduce this load decreasing frames per second will be necessary, This could make the image rendered a bit blurred but this will be OK.

PeterGhimself:
closes #22 

A very simple implementation allowing us to view the arm camera stream, given that the `StreamDispatcher.py` process is running on the rover.
The methods/libraries may have to be changed since mjpeg streams aren't generally supported past the`img` tag and using the newer HTML `video` tag would give us more to work with.
This implementation does _not_ implement starting the stream or stopping it via commands sent from the base station, nor updating FPs. The stream must be initiated manually from the rover itself.

Tested on a single server/client ubuntu 16.04 machine.

proof-issue-22

### DIFF
--- a/robot/basestation/templates/AsimovOperation.html
+++ b/robot/basestation/templates/AsimovOperation.html
@@ -65,7 +65,7 @@
                 </div>
                 <div class="col-md-6">
                     <h5 class="container-header">Arm Vision</h5>
-                    <img src="http://192.168.2.11:8090/?action=stream" class="img-fluid" alt="Responsive image"
+                    <img src="http://{{ roverIP }}:8090/?action=stream" class="img-fluid" alt="Responsive image"
                     />
                 </div>
             </div>
--- a/robot/basestation/templates/AsimovOperation.html
+++ b/robot/basestation/templates/AsimovOperation.html
@@ -33,7 +33,7 @@
         <a class="navbar-brand" href="#">
             <img src="https://spaceconcordia.github.io/img/sclogo_header.png">
         </a>
-        <button class="navbar-toggler" type="button" data-toggle="collapse" d ata-target="#navbarsExampleDefault" aria-controls="navbarsExampleDefault"
+        <button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#navbarsExampleDefault" aria-controls="navbarsExampleDefault"
             aria-expanded="false" aria-label="Toggle navigation">
             <span class="navbar-toggler-icon"></span>
         </button>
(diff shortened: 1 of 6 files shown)


robot/rover/README.md
```diff
@@ -12,6 +15,14 @@ This method installs it in your arduino libraries folder.
 
 If you wish to refer the `include` statements to set your relative path to the libraries, you must surround the libary name with double quotes.
 
+## StreamDispatcher.py
+Currently this starts a stream via `start_stream.sh` which uses mjpgstreamer, more details are documented in the bash script itself.
+To start the stream, assuming you already have a usb web camera connected, run `./StreamDispatcher.py`. Use `Ctrl + C` to stop the stream.
+
+Assuming the stream is being dispatched, it can be accessed in any browser by visiting: `server_ip:8090/?action=stream`, where `server_ip` is replaced with the actual ip address.
```
vashmata: here you mean the ip address of the machine which is streaming the webcam, right?

PeterGhimself: yes, that would be the server.

robot/rover/README.md
```diff
@@ -12,6 +15,14 @@ This method installs it in your arduino libraries folder.
...
+If you have started the StreamDispatcher, opening the Arm page will show the live of feed of the camera in the Arm Vision panel.
+The rover ip address in `app.py` will have to be set to the correct value.
```
vashmata: I guess this answers that question.

robot/rover/README.md
vashmata:
```suggestion
Assuming the stream is being dispatched, it can be accessed in any browser by visiting: `server_ip:8090/?action=stream`, where `server_ip` is replaced with the actual ip address of the server doing the streaming.
```

PeterGhimself: It doesn't have to be the rover. You can test all this with only one machine serving as both client and server, not necessarily the rover.  So I would add "of the server..." but remove the (the rover in this case). Fair?

robot/basestation/templates/AsimovOperation.html
```diff
@@ -65,8 +65,8 @@ <h5 class="container-header">Arm Postion</h5>
                 </div>
                 <div class="col-md-6">
                     <h5 class="container-header">Arm Vision</h5>
-                    <img src="http://{{ roverIP }}:8090/?action=stream" class="img-fluid" alt="Responsive image"
-                    />
+                    <!-- It would be preferable to use video tag, but mjpeg streams are not have no general support beyond img tags -->
+                    <img src="http://{{ roverIP }}:8090/?action=stream" class="img-fluid" alt="Arm Camera Stream Offline"/>
```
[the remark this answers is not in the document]
PeterGhimself: look up flask template vairables

robot/rover/README.md
PeterGhimself: Done, thanks

robot/basestation/templates/AsimovOperation.html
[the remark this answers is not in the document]
PeterGhimself: but basically the template (or view) can be passed variables (in app.py)

Issue #72, opened 2018-12-02
[OBC1] Odroid bootup ping acknowledgement service


PeterGhimself:
On bootup, the Odroid should start a listener process that waits for incoming ping acknowledgement request from the base station and responds to it.

- [x] Package for test client and server nodes created
- [x] Setup ros-server on bootup
- - [x] Configure network interfaces to automatically select `RoverOBC` static IP `192.168.1.30` config over `WiredConnection1` dynamic ip config depending on if `8.8.8.8` is pingable on the auto ethernet config `WiredConnection1`
- - [x] Install ROS and test client/server nodes on odroid locally
- - [x] Configure ROS between basestation and rover and test pinging basestation to rover
- - [x] Set up startup service for starting roscore and ping server node
- - [x] Test pinging from basestation over RocketM900s to OBC
- [ ] Save first prod image and upload to drive

PR #81, opened 2019-01-03
Basestation send ping 74

### LINKED ISSUE #31: Users shall be able to test the connection via ping
Use cases:
- 4.1.19 Ping rover
This is used to determine if communication with rover arm can be established over the network. This will ensure that the GUI can receive the data it needs over the network. It can also be used to test the connection strength and help make judgment calls such as reducing FPS of a certain camera stream because of latency or turning off color of the camera stream to increase overall system performance if necessary.

### LINKED ISSUE #74: [BS1] Basestation send acknowledgement request
Send request to confirm if Odroid is up, running, and ready to receive signals from the basestation.

- [x] Add button on GUI for sending the request
- - [x] Select master IP value based from `ROS_MASTER_URI` environment variable
- - [x] Use bash ping to check if we ping the rover
- - [x] Use ROS service from OBC1 ( #72 ) to request a response. This will let us know if ROS config is good.
- [x] Have response appear in log console
- [x] Add timestamps to ping_acknowledgment package
- [x] Add keyboard short cut `Ctrl + Alt + P`
- [x] Handle error cases
- - [x] Detect/report no network connection on basestation
- - [x] Detect/report no network connection on rover
- [x] Final test with odroid

PeterGhimself:
closes #74 
closes #31

## Summary (check list from issue)
- [x] Add button on GUI for sending the request
- - [x] Select master IP value based from `ROS_MASTER_URI` environment variable
- - [x] Use bash ping to check if we ping the rover
- - [x] Use ROS service from OBC1 ( #72 ) to request a response. This will let us know if ROS config is good.
- [x] Have response appear in log console
- [x] Add timestamps to ping_acknowledgment package
- [x] Add keyboard short cut `Ctrl + Alt + P`
- [x] Handle error cases
- - [x] Detect/report no network connection on basestation
- - [x] Detect/report no network connection on rover
- [x] Final test with odroid

A quick note, with the addition of automatically selecting ros master IP from `ROS_MASTER_URI` environment variable, the ping commands will only work if instead of having it set to `localhost` we set it more directly to `127.0.0.1`


robot/basestation/app.py
```diff
@@ -1,168 +1,206 @@
...
+    """Current Landing page, the arm panel"""
+
+    global rover_ip, rover_ip_raw
+
+    # get rover's IP from global bash environment
+    rover_ip = os.environ["ROS_MASTER_URI"]
+    rover_ip = re.findall(r"(.*):\d+", rover_ip)[0]
+    #print("rover_ip: " + rover_ip)
+
+    rover_ip_raw = rover_ip.split("//")[1]
```
tatumalenko: Why are you redefining `rover_ip_raw` on this function if you define it in its outer scope anyways?

robot/basestation/app.py
```diff
@@ -1,168 +1,206 @@
...
 import flask
+from flask import jsonify
 
 app = flask.Flask(__name__)
 
-# set the rover's IP
-# rover static ip (tenatative): 192.168.1.20
-ROVER_IP = "127.31.43.134"
+# get rover's IP from global bash environment
+rover_ip = os.environ["ROS_MASTER_URI"]
```
tatumalenko: I suggest we setup environment variables a little more safely if we are going to need them. My suggestion is this:

Install `direnv` which autoloads environment variables specified in a `.envrc` file. Check [here](https://direnv.net/) for installation and basic usage.

Then, we create two files in the root project:
- `.envrc`: stores the env variables line by line such as `TEST_VAR="test"`.
- `setupvenv`: a simple shell script that allows to be referenced using `source` to start the `activate` script in the `venv/bin` and also initialize the setuptools file and activate the `direnv` tool. Example:
```
# setupvenv
source venv/bin/activate
python setup.py develop
direnv allow
```

This will take care of environment variables for us and not pollute the global profile. 

Every time we need to work on the project we `cd` into the root, and simply activate the environment and the rest of the toolchain by simply executing:
```
source setupvenv
```

robot/basestation/app.py
PeterGhimself: re-setting it in case it was changed in the meantime. Like if the dev was testing with the odroid, then changed to local.

tatumalenko: I'm still not sure I see why it would change during the execution of the code, it smells dirty. But regardless, at the very least why not abstract the fetching and parsing of the ip value in a function and instead call that function which returns `rover_ip, rover_ip_raw` and use it accordingly. I'll make a commit soon to showcase this along with the other comment about setting up the direnv stuff.

README.md
```diff
@@ -103,6 +103,30 @@ To remove the package you just installed using:
...
+
+To use it, in the project root directory, simply source it in your shell as follows:
+```
+source activatevenv
+```
+
+Tip: You can `source activatevenv` whenever you need to:
+- Reinstall modules defined in `requirements-*.txt`
+- Reactivate the `venv` environment
+- Reload newly defined variables in `.envrc` as environment variables
```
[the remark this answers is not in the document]
tatumalenko: That's what `direnv allow .` does, and that's to avoid sourcing the exported env vars that will pollute the namespace once you're done working on the project (i.e. navigate outside the projects root). Otherwise, yah, we could do that if direnv isn't installed. I'll put that in the if not found block.

tatumalenko:
I approve!

Issue #99, opened 2019-05-15
Integrate ROS directly inside of GUI


vashmata:
This issue is continued by issue #105
### Modify webpage to use ROS directly instead of using workarounds
- [x] Refactor app.py, app.js, etc for general control
- - [x] Add battery voltage calc in Teensy code and display it in GUI
- - [x] Implement requests to services (mux selector, serial commands, task handler)
besides serial commands if listener is not active
- - [x] Implement parameters to directly control local vs competition and usb vs uart inside the webpage? 
- - - [x] Arguments to ArmNode.py control usb vs uart
- - - [x] Control usb vs uart in webpage? use parameters? or task_handler args?
- [x] Refactor app.py, arm.js and Arm.html for arm control
- - [x] Rename AsimovOperation.html to Arm.html
- - [x] Activate ArmNode.py with task_handler instead of arm_listener
- - [x] Implement requests to services (ping, etc - generic)
- - [x] Implement publishing to topics (stop, budge, reset, etc - generic)
- - [x] Handle multiple keypresses for simultaneous budging
- - [x] Implement subscribing to topics (listen for arm angles and update the gui - and other messages?)
- - [x] Update manual controls GUI to be more intuitive
- - [x] Add open/closed loop button
- - [x] Add speed adjustment control
- - [x] Implement homing call 
- [x] Implement protections dealing with serial commands and mux channels ex. arm page:
- - [x] listener is off - serial commands can pick any mux channel and send any serial command using generic thing which opens and closes usb port
- - [x] try to turn listener on - will check if on arm channel. if not attempts to go to arm channel, if can't doesn't turn on. if yes attempts to turn listener on
- - [x] listener is on - refuse to go to other channels and send arm commands
### Build a ROS package for interfacing with Arm Teensy
- [x] Finish making arm node to interface correctly between teensy and ROS
- - [x] Publish messages from teensy to relevant topics
- - - [x] Publish any messages to topics
- - - [x] Deal with newline stripping bug
- - - [x] Filter messages and decide where to publish different types
- - - - [x] Figure out teensy comms to make this easier
- - - [x] Decide where to split up angle messages into distinct topics, consider making composite messages for angles/budge commands/move commands etc
- - [x] Subscribe and respond to commands
- - [x] Respond to client requests
- - - [x] Create or find a .srv file which sends a String and responds with a String (the response) and a bool (indicating success or failure)
- - [x] Refine serial comms code
[...]

PR #104, opened 2019-05-24
Multiple cameras 101

### LINKED ISSUE #101: Handle multiple camera streams
- [x] Differentiate between arm/science, rover front, rover back cameras
- [x] Update task_handler service and `start_stream.sh` accordingly to add multiple camera options
- [x] Enforce mutual exclusivity, only one camera stream can be on at a time
- [x] Corresponding UI updates

PeterGhimself:
## Summary
All backend changes here, no frontend updates.
As long as we keep the usb ports for specific cameras, the changes in this PR will allow us to activate multiple different camera streams one at a time. No two streams can be on at the same time.
Additional upgrades were made to the task_handler ros service, as well as the Listener class to achieve this.
Now we can pass args as a third argument in the client calls, as well as query the running status of any task. Extra protection against running multiple "listener" type tasks was added for safety reasons, since we suspect that having multiple listeners opening the same serial port would mess things up for `pyserial`.

Still need to add UI changes but waiting to merge @vashmata 's branch regarding issue #99 first

### DIFF
--- a/robot/rospackages/src/task_handler/srv/HandleTask.srv
+++ b/robot/rospackages/src/task_handler/srv/HandleTask.srv
@@ -1,4 +1,5 @@
 string task
 int64 status
+string args
 ---
 string response
--- a/robot/rospackages/src/CMakeLists.txt
+++ b/robot/rospackages/src/CMakeLists.txt
@@ -1,1 +1,1 @@
-C:/opt/ros/kinetic/share/catkin/cmake/toplevel.cmake
+/opt/ros/kinetic/share/catkin/cmake/toplevel.cmake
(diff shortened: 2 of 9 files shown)


robot/rospackages/src/task_handler/srv/HandleTask.srv
```diff
@@ -1,4 +1,5 @@
 string task
 int64 status
+string args
 ---
 string response
```
vashmata: it might be cool to have something a bit more clear, like it responds with:
{boolean istaskrunning,
boolean succeeded,
string response}
or maybe even includes an argument confirming which camera stream is on.

Issue #51, opened 2018-07-29
Default + customizable themes


PeterGhimself:
Allowing users to select from premade themes and make their own custom themes if they really don't like our given themes.

PeterGhimself:
Would be cool to have this done directly through the UI, and have a configuration file that it reads from to get the configurable style values/writes to when updating the theme.

MartensCedric:
duplicate of #153

PR #115, opened 2019-07-23
GH-51: RGB Theme Pack Introduced

starlocke:
The first commit at fulfilling the GH-51 request for predefined themes.

Themes are simply RGB backgrounds at this time. Elaboration and specifics TBD.


robot/basestation/static/js/themes.js
```diff
@@ -0,0 +1,58 @@
...
+        console.log('Prefer theme: ' + name);
+        var futuredate = new Date();
+        var futuretime = futuredate.getTime();
+        futuretime += 3600 * 1000 * 24 * 365; // Roughly 1 year into the future
+        futuredate.setTime(futuretime);
+        document.cookie = "theme=" + name + "; expires=" + futuredate.toUTCString() + "; path=/";
+        append_css('themes/' + name);
+    }
+
+    function get_cookie(cname) {
```
PeterGhimself: This function is already defined in `helpers.js`, was there any particular reason for redefining this?

starlocke: A) I didn't technically redefine a function - it has a different name - it has the same functionality - it's technically a duplicate.

B) I hadn't noticed the existence of helpers.js at all. Refactoring could be a TODO item for later.

C) This new get_cookie() is scoped only within the context of $(document).ready(), as far as I can tell, so, it's mostly harmless?

PeterGhimself: Ah I didn't notice the snake_case vs camelCase, fair enough. Nonetheless, since it has the same functionality we might as well use the already existing one. Better to clean this up now than to add duplicate code.

PR #146, opened 2019-08-10
Added timeout to requests

### LINKED ISSUE #129: ROS requests generated in the GUI do not time out
Note that this is low priority. ROS was integrated into the GUI via a javascript library called roslibjs. The person in charge of integrating this is not familiar enough with javascript and jquery to know how to handle timeouts. The result of this is that whenever the GUI generates a request, if there is no response then there will be no feedback in the GUI. What /should/ occur is that after a certain timeout, say 75ms, the GUI should indicate that no response has come from ROS.

MartensCedric:
closes #129

@PeterGhimself Peter said it was fine if everything was 3 seconds.

Now requests will timeout after a specified delay (default 3 seconds) and call the failure callback with a timeout message. I also added to many request an actual callback that prints it to console (instead of doing nothing)


MartensCedric:
Interesting, I'll try to reproduce the issue. Hold on

MartensCedric:
After a short investigation. I was not able to get any errors on the rover or arm page. Although, by going on the science page. I got an error. I had not tested my branch on the science page since I kind of forget it existed, but I will make sure it works on that page too.

This error is not the same error as you describe though. @PeterGhimself can you provide steps to reproduce your problem?

PeterGhimself:
I'm using bash aliases defined in `robot/basestation/config/.bash_aliases`. I just manually merged it into my own `bash_aliases` in my home directory.

Steps to reproduce:
1. Run `rosgui`
2. Run `guistart`
3. Run `rosrun mux_selector mux_selector_server.py local`
4. Run `rosrun task_handler task_handler_server.py`
5. Visit `localhost:5000`
6. Press `F12` to see errors
7. Change mux channel

MartensCedric:
Ready for a new review

PR #200, opened 2019-10-19
Programming Guidelines

MartensCedric:
For a formatted view : https://github.com/space-concordia-robotics/robotics-prototype/blob/guidelines-178/CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -187,7 +187,7 @@ We use the PEP8 Style Guide. Please refer to it [here](https://www.python.org/de
 
 We follow the [standard](https://github.com/standard/standard/blob/master/RULES.md) conventions, there exists an auto-formatter package in atom called `atom-standard-formatter` which automates this.
 
-Furthermore, we are attempting to use a strict variant of JavaScript. We have not enable `strict mode` yet, but all future code should respect the strict variant of JavaScript which can be found [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode)
+Furthermore, we are attempting to use a strict variant of JavaScript. We have not enabled `strict mode` yet, but all future code should respect the strict variant of JavaScript which can be found [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode)
 
 We are also using JQuery, therefore you should use JQuery for document traversal and and manipulation. 
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -151,7 +151,7 @@ Now it is much cleaner when the call is being read.
 
 ```
 //some code
-setAccelerationLimit(false)
+setAccelerationLimit(false);
 //more code
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,17 +19,19 @@ Teams that have moved swiftly at the beginning of projects can find themselves m
 
 ## Best Practices
 
-### Functions
+These best practices are mostly taken from the book `Clean Code, A Handbook of Agile Craftsmanship` by Robert C. Martin 
 
-#### Do one thing
+### 1. Functions
+
+#### 1.1 Do one thing
 
 Functions should do one thing. They should do it well. They should do it only. One symptom of doing more than one thing is having sections within functions.
 
-#### Size
+#### 1.2 Size
 
-Size does matter! The size of the function should be small. A function that cannot be read in one screen should definitely be split up.
+Size does matter! The size of the function should be **small**. A function that cannot be read in one screen should definitely be split up.
 
-#### Duplication
+#### 1.3 Duplication
 
 Duplicating code is one of the primary source of hard to maintain code. Here is an example of duplicated code in the code base of robotics-prototype
 
@@ -63,30 +65,20 @@ What we gain by eliminating duplication is the need to change the function multi
 There are more changes we could do to improve this function that we will see later.
 
 
-### Coupling
-<<<<<<< HEAD
-
-#### Modularity
-
-#### Levels of abstraction
-
-### Comments
-
-#### Rotted Comments
-=======
+### 2. Coupling
 Coupling is the degree of interdependence between software modules.
 
-#### Modularity
+#### 2.1 Modularity
 Since the rover needs to be modular, we need to have very loose coupling. The software needs to have independent modules that we can quickly swap out. This is one of the many reasons we use ROS, which creates a modular foundation.
 
-#### Levels of abstraction
-it is a best practice to abstract the code into different layers. For example, code related to communication with ROS should not include GUI changes. Common layers are GUI, Logic, Communication.
+#### 2.2 Levels of abstraction
+It is a best practice to abstract the code into different layers. For example, code related to communication with ROS should not include GUI changes. Common layers are GUI, Logic, Communication.
 
 
-### Comments
+### 3. Comments
 Comments can be a blessing or a cure. Proper documentation can help developers while bad documentation can obstruct or mislead them.
 
-#### Useless comments
+#### 3.1 Useless comments
 Ideal code is self explanatory. Do not comment when it is obvious like
 
 ```
@@ -96,11 +88,14 @@ initRos();
 
 This adds lines for the programmers to read without adding any value.
 
-#### Rotted Comments
-One major issue with comments is that they rot. Before writing a comment, make sure that this comment will still be relevant after the code passes through a lot of changes.
->>>>>>> f11aa33... Add section on comments
+#### 3.2 Rotted Comments
+One major issue with comments is that they rot with time. Before writing a comment, make sure that this comment will still be relevant after the code passes through a lot of changes.
 
-### Meaningful names
+#### 3.3 Todos
+Do not write any `Todo` comments. These get easily lost and forgotten. Open an issue instead.
+ 
+
+### 4. Meaningful names
 
 With the function that we improved above there is still a glaring issue :
 
@@ -112,7 +107,7 @@ void Commands::setTto(bool throttleTimeOut) {
 }
 ```
 
-What the heck is `TTO`? Observing this function and argument, it's obvious that it's for `Throttle Timeout`. However, imagine that you were stumbling upon a call to this function like this
+What the heck is `TTO`? By observing this function and it's argument, it is obvious that it's for `Throttle Timeout`. However, imagine that you were stumbling upon a call to this function like this
 
 ```
 //some code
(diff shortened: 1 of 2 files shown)


MartensCedric:
I invite everyone to read through these guidelines and make suggestions. These guidelines will define what we accept and refuse in pull requests.

CONTRIBUTING.md
```diff
@@ -0,0 +1,193 @@
...
+
+### 1. Functions
+
+#### 1.1 Do one thing
+
+Functions should do one thing. They should do it well. They should do it only. One symptom of doing more than one thing is having sections within functions. These can oftentimes easily be split up into subroutines that will only have a specific role.
+
+#### 1.2 Size
+
+Size does matter! The size of the function should be **small**. A function that cannot be read in one screen should definitely be split up.
```
PeterGhimself: What if it is my phone screen? I think it's a little too open to interpretation

CONTRIBUTING.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### 2.1 Modularity
+Since the rover needs to be modular, we need to have very loose coupling. The software needs to have independent modules that we can quickly swap out. This is one of the many reasons we use ROS, which creates a modular foundation.
+
+#### 2.2 Levels of abstraction
+It is a best practice to abstract the code into different layers. For example, code related to communication with ROS should not include GUI changes. Common layers are GUI, Logic, Communication.
+
+
+### 3. Comments
+Comments can be a blessing or a cure. Proper documentation can help developers while bad documentation can obstruct or mislead them.
```
PeterGhimself: blessing or a _curse_

CONTRIBUTING.md
```diff
@@ -0,0 +1,193 @@
...
+Ideal code is self explanatory. Do not comment when it is obvious like
+
+```
+//Initialize ROS
+initRos();
+```
+
+This adds lines for the programmers to read without adding any value.
+
+#### 3.2 Rotted Comments
```
PeterGhimself: I think rotted is used as a verb, so maybe rotten (adj) would be better?

CONTRIBUTING.md
```diff
@@ -0,0 +1,193 @@
...
+One major issue with comments is that they rot with time. Before writing a comment, make sure that this comment will still be relevant after the code passes through a lot of changes.
+
+#### 3.3 Todos
```
PeterGhimself: Sometimes while working on an issue I will leave TODOs for myself to later quickly control-f to that location, for instance. That being said, this is meant as a PR guideline so it makes sense to keep those out of master.

PeterGhimself: Opening an issue would not apply if the `TODO` is really trivial... I guess that's implied but maybe should be made more clear.

CONTRIBUTING.md
```diff
@@ -0,0 +1,193 @@
...
+//some code
+setTto(false);
+//more code
+```
+To know what `TTO` means, you need to go to the function definition which could be tedious if you have to do this often.
+
+Furthermore, you would expect `throttleTimeOut` to be a number, the timeout time, but it is a boolean.
+
+```
+void Commands::hasThrottleTimeout(bool hasThrottleTimeout) {
```
PeterGhimself: should the function be named with `has` at the front? It does not return a boolean, just basically prints text based off the value of a given boolean.

CONTRIBUTING.md
```diff
@@ -0,0 +1,193 @@
...
+
+Magic values should be avoided and replaced by variables or constants with names that self-describes their behavior.
+
+
+
+## Programming Specific Guidelines
+
+### Python
+
+We use the PEP8 Style Guide. Please refer to it [here](https://www.python.org/dev/peps/pep-0008/)
```
PeterGhimself: Did you take [this](https://github.com/space-concordia-robotics/robotics-prototype/wiki/Code-Formatting-and-Conventions#formatting-guide-python) into consideration? We have this already in place for enforcing python programming style and standards using autoformatter/linter combo in `requirements-dev.txt`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Best Practices
+
+These best practices are mostly taken from the book `Clean Code, A Handbook of Agile Craftsmanship` by Robert C. Martin 
+
+### 1. Functions
+
+#### 1.1 Do one thing
+
+Functions should do one thing. They should do it well. They should do it only. One symptom of doing more than one thing is having sections within functions. These can oftentimes easily be split up into subroutines that will only have a specific role.
```
PeterGhimself: "They should do it alone" sounds bit better to me here than "They should do it only" but I think either works to get the point across.

PeterGhimself: Additionally, 1.1 only applies to functions in this document it could be extended to classes as well as per the [swiss army knife anti-pattern](https://sourcemaking.com/antipatterns/swiss-army-knife)

CONTRIBUTING.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Programming Specific Guidelines
+
+### Python
+
+We use the PEP8 Style Guide. Please refer to it [here](https://www.python.org/dev/peps/pep-0008/)
+
+### JavaScript
+
+We are attempting to use a strict variant of JavaScript. We have not enable `strict mode` yet, but all future code should respect the strict variant of JavaScript which can be found [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode)
```
PeterGhimself: `strict mode` was introduced in ECMAScript5 but we try to adhere to ECMAScript6 as per the first line [here](https://github.com/space-concordia-robotics/robotics-prototype#contributing-and-development-environment-instructions)

CONTRIBUTING.md
MartensCedric: https://wikidiff.com/rotten/rotted

The former is what I want to express. "overridden with bacteria and other infectious agents"

CONTRIBUTING.md
MartensCedric: I've appended a few sentences to the TODO paragraph. Please re-read it. This is meant to include what I think is a good resolution. That TODO's should just not be present on PR's, but it's totally fine that you commit stuff with TODO's in it, as along as they have either opened issues or are actually addressed. What do you think?

CONTRIBUTING.md
MartensCedric: Ah nevermind, I have one IQ right now. I didn't even realize it was Rotten I needed

CONTRIBUTING.md
MartensCedric: So I don't want to put a hard limit to the number of lines of code. I updated the paragraph with information from this https://softwareengineering.stackexchange.com/a/133406/255377

CONTRIBUTING.md
MartensCedric: true, it sets something though

CONTRIBUTING.md
MartensCedric: > "They should do it alone" sounds bit better to me here than "They should do it only" but I think either works to get the point across.

It's actually a quote from a software engineering book on clean code that I copy pasted

MartensCedric: Good point with this antipattern

CONTRIBUTING.md
PeterGhimself: LGTM

CONTRIBUTING.md
```diff
@@ -25,42 +25,52 @@ These best practices are mostly taken from the book `Clean Code, A Handbook of A
...
 
-Functions should do one thing. They should do it well. They should do it only. One symptom of doing more than one thing is having sections within functions. These can oftentimes easily be split up into subroutines that will only have a specific role.
+Functions should do one thing. They should do it well. They should do it only. This is called the Single Responsibility Principle (SRP). One symptom of doing more than one thing is having sections within functions. These can oftentimes easily be split up into subroutines that will only have a specific role. Additionally, this software engineering design principle extends to classes. Classes should have a clear purpose and not attempt to be a Swiss Army Knife.
 
 #### 1.2 Size
 
-Size does matter! The size of the function should be **small**. A function that cannot be read in one screen should definitely be split up.
+Size does matter! The size of the function should be **small**. We do not impose a hard limit on the number of lines of code. However, common sense should be applied. Code should be able to be read on a single monitor. It should be short enough to be able to conceptualize the logc in your head. Use common sense.
```
PeterGhimself: _logic_ instead of _logc_

CONTRIBUTING.md
```diff
@@ -105,40 +115,45 @@ Avoid commenting code. A lot of the code that is commented, never gets uncomment
...
 ```
 
-Boolean variable should start with `has` or `is`.
-
+Now it is much cleaner when the call is being read.
 
 ```
 //some code
-hasThrottleTimeout(false);
+setAccelerationLimit(false)
```
PeterGhimself: Missing semi-colon, though it is just for example sake we might as well keep it syntactically correct.

MartensCedric: Yup will fix, got used to JS and Python too much

CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+        motorList[i].accLimit = hasAccLimit;
+        String msg = "ASTRO Motor " + String(i + 1);
+        msg += " Acceleration Limiter: ";
+        msg += String(motorList[i].accLimit ? "Open" : "CLose");
+        println(msg);
+    }
+}
+```
+
+Now it is much cleaner when the call is being read.
```
PeterGhimself: self-documenting-regex

CONTRIBUTING.md
MartensCedric: there's a key in the yapf file that we can put `based_on_style = pep8`. Should we do it ? Maybe discuss IRL

CONTRIBUTING.md
MartensCedric: is this gucci now?

CONTRIBUTING.md
MartensCedric: I will also discuss that we use Standard JS from here as explained on the Wiki https://github.com/standard/standard/blob/master/RULES.md

MartensCedric: done

CONTRIBUTING.md
PeterGhimself: yeah might as well do it

CONTRIBUTING.md
PeterGhimself: gucci

CONTRIBUTING.md
PeterGhimself: yeah this is what we try to enforce with the `standard-formatter` in atom

CONTRIBUTING.md
```diff
@@ -185,7 +185,9 @@ We use the PEP8 Style Guide. Please refer to it [here](https://www.python.org/de
 
 ### JavaScript
 
-We are attempting to use a strict variant of JavaScript. We have not enable `strict mode` yet, but all future code should respect the strict variant of JavaScript which can be found [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode)
+We follow the [standard](https://github.com/standard/standard/blob/master/RULES.md) conventions, there exists an auto-formatter package in atom called `atom-standard-formatter` which automates this.
+
+Furthermore, we are attempting to use a strict variant of JavaScript. We have not enable `strict mode` yet, but all future code should respect the strict variant of JavaScript which can be found [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode)
```
PeterGhimself: We have not _enabled_

MartensCedric: done

CONTRIBUTING.md
MartensCedric: done

PR #205, opened 2019-11-02
Save camera feed 23

### LINKED ISSUE #23: Save camera feed
- [x] Add method for saving stream
- [x] Modify GUI to add buttons for saving stream

Use cases:
- 4.1.24 Enable record camera feed
By pressing the "record feed" toggle button to set if ON,, the stream being captured by the base-station will be saved into a predefined folder.
- 4.1.25 Disable record camera feed
By pressing the "record feed" toggle button to set if OFF, the recording of the camera feed will cease. This should be the default state.

NinjaWeeaboo:
Closes #23


MartensCedric:
Some things that I discussed with @NinjaWeeaboo 

1. Currently it does not fully support multiple camera stream (Does this fall into the scope of the multiple camera stream task)

robot/basestation/app.py
```diff
@@ -245,6 +247,36 @@ def capture_image():
 
     return jsonify(msg=msg)
 
+# This is called on if ffmpeg isn't already recording when the record toggle button is clicked
```
MartensCedric: I know we aren't utilizing this much but we should adhere to the PEP8 standards that we are trying to follow. The way to document stuff in python is like this https://www.python.org/dev/peps/pep-0008/#documentation-strings. You can learn more about them here : https://www.pythonforbeginners.com/basics/python-docstrings 

Additionally, I would not mention "when the record toggle button is clicked" because perhaps we would reuse that function from another source eventually.

robot/basestation/app.py
```diff
@@ -245,6 +247,36 @@ def capture_image():
...
+@app.route("/start_recording_feed/<stream>", methods=["POST", "GET"])
+def start_recording_feed(stream):
+    date = datetime.datetime.now().strftime("%I:%M:%S_%B_%d_%Y")
+    # Name of file to be saved
+    name = stream + '_' + date
+    # Get stream url
+    stream_url = "http://" + fetch_ros_master_ip() + ":8080/stream?topic=/cv_camera/image_raw"
+    # Message to print to log
```
MartensCedric: I would recommend changing your variables to something more descriptive and remove the comments. 

example

`formatted_date = ...`

`name_of_file = ...`

`recording_log_msg`

Right now many comments must be read because the variable names are not descriptive enough

robot/basestation/static/js/elements/feed/camera.js
```diff
@@ -112,6 +112,39 @@ $(document).ready(() => {
     })
   })
 
+  $('#record-video-toggle').on('click', function (event) {
+    var el = $(this); // variable of button text
```
MartensCedric: Use `let` instead of `var`.  https://www.geeksforgeeks.org/difference-between-var-and-let-in-javascript/

Here it is a good instance of "Making good variable names to remove comments`

This line should be something `let recordingButton = $(this);`
There is no need to comment since it is obvious. You can look into the rest of your code to see what can be improved like this.

robot/basestation/static/js/elements/feed/camera.js
```diff
@@ -112,6 +112,39 @@ $(document).ready(() => {
...
+
+    // Swap button text out for 'Start recording feed' if button says 'Stop recording feed'
+    // run stop recording feed function in app.py
+    if (el.text() == el.data("text-swap")) {
+        el.text(el.data("text-original"));
+        el.css({ 'font-weight': 'bold' }) // Change font back to bold
+
+        // Call on stop_recording_feed function in app.py
+        $.ajax('/stop_recording_feed/default', {
```
MartensCedric: What if the ajax function fails? Then the button will still switch.

MartensCedric: It should maybe also log an error

robot/basestation/app.py
```diff
@@ -245,6 +247,36 @@ def capture_image():
 
     return jsonify(msg=msg)
 
+# This is called on if ffmpeg isn't already recording when the record toggle button is clicked
+@app.route("/start_recording_feed/<stream>", methods=["POST", "GET"])
+def start_recording_feed(stream):
```
MartensCedric: This function has many responsibilities. I would recommend making another function about starting and stopping ffmpeg.

robot/basestation/static/js/elements/feed/camera.js
```diff
@@ -112,6 +112,39 @@ $(document).ready(() => {
     })
   })
 
+  $('#record-video-toggle').on('click', function (event) {
+    var el = $(this); // variable of button text
+
+    // Swap button text out for 'Start recording feed' if button says 'Stop recording feed'
+    // run stop recording feed function in app.py
+    if (el.text() == el.data("text-swap")) {
```
MartensCedric: This may be a bit more advanced, but I foresee the need to have other "toggle-able" buttons. So having this logic in every of them would not be optimal. Instead there are elegant solutions in this post https://stackoverflow.com/questions/2155453/jquery-toggle-text

robot/basestation/app.py
NinjaWeeaboo: Aren't they already separate functions? Or do you want an seperate function from the one that makes variables, that would handle starting the stream?

NinjaWeeaboo:
I'm ready to be cooked alive from the roasts on my code.

NinjaWeeaboo:
BTW, I figured out a way to handle the error caused by cv_camera not running.

robot/basestation/static/js/elements/tools/error_handling.js
```diff
@@ -0,0 +1,18 @@
+function flaskError (jqXHR, exception) {
```
MartensCedric: I would suggest you slightly expand this function with a third argument on the URL of the error.

Suppose we have multiple systems that use your flaskError function then if we see "error 404" we don't actually know what subsystem/action triggered it. 

So I would like to see a message like "Error 404 when accessing localhost:5000/rover" or something

MartensCedric:
So one issue that arises is if you start a stream and you refresh, it will put it automatically back to "start" which shouldn't be the case. It will continue recording until flask is killed.

Not sure if it should be fixed here or not @PeterGhimself . I think there's already been much more validation/error handling done in this PR than anything yet. we could open an issue up for that one specifically, the fix doesn't add that much value and we could give it as a new member task to fix that.

NinjaWeeaboo:
Thanks! I can probably fix the issue with the button if you guys want. Although I guess it would make a good first issue task, but that's up to you guys.

NinjaWeeaboo:
It was a really quick fix to stop the button from being reset back to 'start' upon refreshing the page. So I just took care of it.

robot/basestation/stream_capture.py
```diff
@@ -0,0 +1,48 @@
...
+    save_video_dir = 'rover_stream/' + stream
+    stream_url = "http://" + fetch_ros_master_ip() + ":8080/stream?topic=/cv_camera/image_raw"
+    return feed_connection_check(stream, filename, save_video_dir, stream_url, recording_log_msg)
+
+def stop_recording_feed(stream):
+    recording_log_msg = 'saved recording of ' + stream
+    proc_video[stream].communicate(b'q')
+    return jsonify(recording_log_msg=recording_log_msg)
+
+def feed_connection_check(stream, filename, save_video_dir, stream_url, recording_log_msg):
```
PeterGhimself: I feel like this function has a bit too many parameters, like `save_video_dir`, could that not be a constant?
As for `stream` and `stream_url` could `stream_url` be defined based of `stream`?

robot/basestation/stream_capture.py
```diff
@@ -0,0 +1,48 @@
+import subprocess
```
PeterGhimself: All the newly added functions should have [docstrings](https://www.datacamp.com/community/tutorials/docstrings-python)

MartensCedric:
I shall run the GIT

MartensCedric:
Successfully Passes GIT

PR #210, opened 2019-11-03
Jsload 206

### LINKED ISSUE #206: JS load error
The problem is most likely some dependencies between different js files and the fact that even listeners only get bound properly to html elements when they are loaded in a certain order.

To reproduce:

- Checkout [this](https://github.com/space-concordia-robotics/robotics-prototype/tree/8a7f01de909f3555d7d5fd3a57895a98974f2a7c) version of the repo
- Run gui with `startgui` and go to rover page at `localhost:5000/rover`
- Click any of the extra controls buttons and they will not work

If you remove the first and last line (the `$(document).ready` part) it will work again. This is only a temporary hack/fix.

MartensCedric:
Closes #206


MartensCedric:
> Tested and seems to work fine except on the science page, can we look into this?

Yup, it was missing an js import. The same problem was also on the PDS page. Fixed

PR #232, opened 2019-12-06
[#220] Clarify README and fix cv_camera issues

### LINKED ISSUE #220: Can't build cv_camera on a new setup
Following the current readme will lead to an error while building cv_camera. The fix is to require new users to install some dependencies that contain the .cmake files.

MartensCedric:
Closes #220


MartensCedric:
One step that was omitted that you will need to follow @JonathanSolvesProblems. If this is not merged by the 17th make sure you follow the extra step added

AFknUsername:
2fast

Issue #201, opened 2019-10-19
Camera template that includes buttons


MartensCedric:
Modify the camera template to include these : 

## Required

- [x] Rotation
- [x] Label (Name)
- [x] Dropdown to select stream (valid options: `/dev/videoArmScience`, `/dev/videoFront`, `/dev/videoRear`)
- [x] Power button
- [x] Take Screenshot
- [x] Stop/Start recording
- [x] Controls overlay

## Optional

- [ ] Make dropdown port option default (only initially) to respective competition mode values based off page

## Future considerations 
- [ ] Change background color of each option in dropdown list based off if the options would be valid (like if we're in local mode, which would be checkable by what `ROS_MASTER` is set to, then we can assume that the competion mode values are not valid from the basestation perspective)

![image](https://user-images.githubusercontent.com/18620025/67151561-e2c32700-f295-11e9-9ba3-9ec0e8e0f3a1.png)

MartensCedric:
So we should add a dropdown to select camera port stream

MartensCedric:
![Concept](https://user-images.githubusercontent.com/18620025/71301531-fcecb480-236d-11ea-94fc-971fab015668.png)

Concept Art by @Kayram2710

MartensCedric:
Stuff I may forget that is important to have 

- [ ] Implement an identification system to ensure that the javascript button doesn't trigger all the cameras at the same time
- [ ] rotation should work 360 degrees
- [ ] new stream offline art "Space Concordia Robotics | Stream Offline" @mattsklivas

PR #248, opened 2019-12-21
Simultaneous streams 151 + localized camera stream testing 244

### LINKED ISSUE #151: Simultaneous camera streams backend
Currently the way video streaming is implemented in the code base is that by design we restrict the odroid to only dispatching one stream at a time, so always the same link is used on the client to capture this stream. In fact the current placement of usb cameras does not allow us to stream multiple at the same time, since they are all connected to the same USB hub connected to a single port, there isn't enough bandwidth to stream two over said port, but there may be if we reorganize the usb camera placements with one on the 2.0 hub and another on the 3.0 hub. The default resolution we stream at is 720p (SD). This issue would deal with modifying the backend to allow multiple streams to be on at the same time.
We know that with our ERC radios we get a bandwidth of ~90 Mbps (source: @Waleed-HM ) so it should be able to handle the amount of data being streamed over the rover network. If any issues arise due to this the streaming individual cameras at lower resolution could help mitigate those issues.

- [x] Backend
- - [x] Update task handler and `start_ros_stream.sh` to be able to handle multiple streams at once
- - [x] Dynamic comp/local test mode configuration
- - [x] Test if streaming at lower resolutions allows for multiple streams active on odroid at the same time
- - - [x] 240p --> allows 3 streams max
- - - [x] 360p --> allows 2 streams max, when starting 3rd one we get: `VIDIOC_STREAMON: No space left on device` and the [blue light flicker](https://obsproject.com/forum/threads/logitech-webcam-c920-flickering.95956/)
- - - [x] ~~480p~~ --> allows 1 stream max
- - - [x] ~~720p~~ --> turning on second camera makes the blue light flicker
- - [x] Modify rover usb port configuration for cameras to be able to stream more than one camera
- - [x] Add sub command in task handler for camera stream, new option to get available video usb ports
- - [x] Add support for checking which camera stream ports are currently active
- [x] Address breaking changes to frontend
- - [x] Temporary fix for incorrect URL on frontend, not to enter the scope of #201
- [x] Regression test on rover
- [ ] Optional
- - [ ] Include setting default resolution in `cv_camera` node instantiation
- - [ ] Add feature for request to start/restart node with specific resolution
- - [ ] Front end for start/restart node with specific resolution (dropdown list of reasonable options)
- - [ ] Add response for request which ports are available

### LINKED ISSUE #244: Localized testing of camera streams
## Issue description
Currently the task handler only accepts the following parameters to run streams `/dev/ttyArmScienceCam`, `/dev/ttyFrontCam`, `/dev/ttyRearCam`, which is makes the code simple since it doesn't have to try to figure out which camera it's talking to now. This mapping is achieved by [this file](https://github.com/space-concordia-robotics/robotics-prototype/blob/master/robot/rover/config/my-webcams.rules) which is meant to be run on the odroid OBC.
There should be a way to run this part of the task handler in local mode so that developers could simulate that subsystem without the need to do any system configuration, just running the task handler in local mode like:
- `rosrun task_handler task_handler_server.py local`

## Task Breakdown
- [x] Allow for args to be any valid instance of `/dev/video*` for the computer on which task_handler_server.py runs on
- [x] Add 'local' parameter for server and client task handler scripts to switch between modes
- [ ] Run GIT for regression checks

PeterGhimself:
# Assignee Section

## Description
The changes here towards allowing multiple streams to be run simultaneously through the task handler are purely backend, making ammends for the way #104 was done. Although a few small front end changes are included here, the purpose of this PR is to address the backend and therefore it does not close issue #151 but #244 instead.
So in local mode it actually checks and gets a list of ports available where the task_handler_server script is running that follow the pattern `/dev/video[0-9]*`, whereas in comp mode we only address fixed names `/dev/ttyArmScienceCam`, `/dev/ttyFrontCam`, `/dev/ttyRearCam` since we cannot reliably use the real names of the ports (/dev/video8,7,6, etc.) see [here](https://github.com/space-concordia-robotics/robotics-prototype/tree/master/robot/rover/config#my-webcamsrules) for an explanation on why.

- [x] Task handler allows for multiple streams to be turned ON simultaneously, and turning off specific ones works too
- [x] Task handler client and server scripts accept 'local' argument, camera_stream task supported
- [x] `local_rover.launch` roslaunch file updated

### Example usage:
start simulating rover
- `roslaunch local_rover.launch`

start video streams on webcams 0 and 1 (usually /dev/video0 is mapped to your built in laptop webcam, and everytime you plug in a new usb camera it will assign /dev/video[ctr++] kind of idea)
- `rosrun task_handler task_handler_client.py camera_stream 1 /dev/video0 local`
- `rosrun task_handler task_handler_client.py camera_stream 1 /dev/video1 local`

stop video streams on webcams 0 and 1
- `rosrun task_handler task_handler_client.py camera_stream 0 /dev/video0 local`
- `rosrun task_handler task_handler_client.py camera_stream 0 /dev/video1 local`

You can verify the stream is running by visiting links like:
- http://0.0.0.0:8080/stream_viewer?topic=/video0/image_raw
- http://0.0.0.0:8080/stream_viewer?topic=/video1/image_raw

closes #244
closes #151

## Check list for testing locally
- [x] Turn on/off and view individual streams one by one through CLI
- [x] Turn on multiple streams at once through CLI and view them, control which ones to turn off at a time all through CLI
- - [x] Two cameras
- - [x] Three cameras

## Check list for testing with rover
- [x] Turn on/off and view individual streams one by one through CLI
- [x] Turn on multiple streams at once through CLI and view them, control which ones to turn off at a time all through CLI + GUI
- - [x] Two cameras
- - [x] Three cameras

# Reviewer Section

- [x] [General Integration Test](https://docs.google.com/document/d/1ug0CpA1cIzURP8DDFSvCt2CEJJSwJ6Ta6B1LG_hYk6I/edit) Performed Successfully


PeterGhimself:
I haven't yet run a test on the rover itself for this branch, planning to do this either tonight or tomorrow.

robot/basestation/app.py
```diff
@@ -89,6 +89,12 @@ def pds():
 def numSections():
     return '4'
 
+@app.route("/stream")
+def stream():
+    """Stream page."""
+    return flask.render_template("pages/Stream.html", roverIP=fetch_ros_master_ip())
+
+# we still using this?
```
MartensCedric: delete dis

robot/basestation/static/js/stream.js
```diff
@@ -0,0 +1,33 @@
...
+    $('#rear-stream-btn').on('click', function (event) {
+        // click makes it checked during this time, so trying to enable
+        if ($('#rear-stream-btn').is(':checked')) {
+            $('img#rear-camera-feed')[0].src =
+            'http://localhost:8080/stream?topic=/RearCamera/image_raw'
+        } else {
+            $('img#rear-camera-feed')[0].src =
+                '../static/images/stream-offline.jpg'
+        }
+    })
```
MartensCedric: This is code duplication. I suggest creating a function of this signature

`string getStreamURL(streamName)`
That returns 'http://localhost:8080/stream?topic=/' +streamName +' /image_raw'

Also put the '../static/images/stream-offiline.jpg' in a constant.

then you have at maximum
```
$('#rear-stream-btn).click(function(){
 if ($('#rear-stream-btn').is(':checked')) {
            $('img#rear-camera-feed')[0].src = getStreamURL('RearCamera')
        } else {
            $('img#rear-camera-feed')[0].src = CONSTANT
        }
})
```

I would also remove the comment as it's a bit confusing, reading the code is much simpler. What does it try to say by click makes it checked _during this time_ (What time?), so trying to enable??

robot/rospackages/src/mux_selector/scripts/mux_select_server.py
```diff
@@ -140,9 +140,8 @@ def mux_select_server():
         print('Exiting... Perhaps you mean\'t \'local\'?')
         sys.exit(0)
     if 'local' in non_ros_args:
-        print("Running in local mode")
```
MartensCedric: why? is it because we already have (local mode) ?

robot/rospackages/src/task_handler/scripts/task_handler_client.py
```diff
@@ -40,24 +50,40 @@ def usage():
...
     Return string showcasing proper usage of this client script
     """
-    help_msg = """USAGE:\nrosrun task_handler task_handler_client.py [task] [status] optional:[args]
-                  \nValid task options: ['arm_listener', 'arm_listener', 'science_listener', 'camera_stream']
-                  \nValid status options: [0, 1, 2]
-                  \nValid camera stream args: ['/dev/ttyFrontCam', '/dev/ttyRearCam', '/dev/ttyArmScienceCam']"""
+    global ports
+
+    help_msg = "USAGE:\nrosrun task_handler task_handler_client.py [task] [status] optional:[args]"
+    help_msg += "\nValid task options: ['arm_listener', 'arm_listener', 'science_listener', 'camera_stream']"
```
MartensCedric: why is the `arm_listener` valid twice lol?

You should probably make a string array of valid args and loop it here and reuse it when checking the args so its synced with the help

MartensCedric: like I mean use the array known_tasks

robot/rospackages/src/task_handler/scripts/task_handler_client.py
```diff
@@ -40,24 +50,40 @@ def usage():
...
     rospy.init_node("task_handler_client")
 
     task = sys.argv[1]
     status = int(sys.argv[2])
 
-    if len(sys.argv) == 4:
+    if len(sys.argv) >= 4:
         args = sys.argv[3]
     else:
         args = ""
```
MartensCedric: what is this for

robot/rospackages/src/task_handler/scripts/task_handler_server.py
```diff
@@ -106,10 +161,50 @@ def handle_task(task, status, args):
...
+    known_listeners = known_tasks[:-1]
+
+    # keep track of currently running streams
+    active_ports = []
+    active_streams = []
+
+    i = 0
+    for task in running_tasks:
+        task.set_name(known_tasks[i])
+        i += 1
```
MartensCedric: for i in range(len(running_tasks)):
    task = running_tasks[i]
    task.............

This makes sure that the scope of the iteration variable is correct

robot/rospackages/src/mux_selector/scripts/mux_select_server.py
PeterGhimself: exactly

robot/rospackages/src/task_handler/scripts/task_handler_client.py
PeterGhimself: yeah it's a typo it one of them should be `rover_listener`
I'll construct the string by using the `known_tasks` array

robot/basestation/app.py
PeterGhimself: done, and the answer is: yes we still need this

robot/basestation/static/js/stream.js
PeterGhimself: Agreed, fixed

robot/rospackages/src/task_handler/scripts/task_handler_client.py
PeterGhimself: I think it was meant to cause an error to be thrown by the following `is_valid_request()` check.
I have removed the `else` block since the effect is the same.

robot/rospackages/src/task_handler/scripts/task_handler_server.py
PeterGhimself: Fixed, thanks

robot/rospackages/src/task_handler/scripts/task_handler_client.py
PeterGhimself: Fixed

PeterGhimself:
So I ran the GIT more or less and nothing broke expect for what is expected: front end for viewing camera streams doesn't work. However, pressing the start stream toggle buttons does turn the cameras on (just the GUI doesn't properly acknowledge that by keeping the toggle ON).
Unfortunately with our current setup it seems the odroid can't handle streaming two cameras at once. I'm assuming this is the case since when I already have one camera running and turn on one of our logitech c920 it starts flickering it's status light, which according to [this](https://obsproject.com/forum/threads/logitech-webcam-c920-flickering.95956/) can mean a few things and I think it's most likely the issue of not having enough bandwidth.
As @KoaTheKoala pointed out before we may be able to stream multiple cameras if they are connected to the 3.0 USB hub, but right now the usb hub we are using for the cameras is connected to the sole 2.0 USB port.
So moving forward I'm going to attempt to:
- [ ] Reorganize the placement of usb hub in a way such that we can stream multiple cameras
- [ ] Re-map usb port aliases accordingly
- [ ] Change the aliases from `/dev/ttyArmScienceCam` to `/dev/videoArmScienceCam` 
- [ ] Modify the logic for determining max amount of streams possible since it works well for the local mode but using /dev/tty* as a pattern on the OBC is very misleading (we get a maximum of over 200 streams this way)

robot/rospackages/src/task_handler/scripts/task_handler_server.py
```diff
@@ -2,33 +2,26 @@
...
 
+# process task, status and corresponding args sent from client
+# return response message
 def handle_task(task, status, args):
-
     response = "Nothing happened"
 
     global running_tasks
+    global stream_ctr
+    global local
```
Davidster: Global variables monkaS

MartensCedric: yes monkaS

MartensCedric: and what is local, do you mean isLocal? also is stream_ctr could be seen as a temp var it's not clean from the name if its the total count of the streams or not

MartensCedric: When you look at the function `handle_task(task, status, args)`, ok that's fine. But I don't think it should be having variables like local and stream_ctr. It's not its responsibility.

if chosen_task is "camera_stream" then it should call a function 

handle_cam_stream_task or something

robot/rospackages/src/task_handler/scripts/task_handler_server.py
```diff
@@ -106,10 +161,48 @@ def handle_task(task, status, args):
     return response + "\n"
 
 def task_handler_server():
+    global local
```
MartensCedric: I think that running it locally should be an argument, also may allow us to call this function externally eventually  for testing using an argument

robot/rospackages/src/task_handler/scripts/task_handler_server.py
PeterGhimself: So far I have
- [x] changed `stream_ctr` to `active_stream_ctr`
- [x] changed `local` to `is_local`
- [x] refactored parts of `handle_task` into functions:
- - [x] validate_camera_stream
- - [x] start_camera_task
- - [ ] stop_camera_task

I'm not sure if the global variables can be avoided at this point since multiple functions need them. I suppose making it a class with those global variables as static fields would be a solution. Would it be fine to open a new issue for turning it into a class? There might be other implications that are going to cause it to be a rabbit hole in itself to implement.

robot/rospackages/src/task_handler/scripts/task_handler_server.py
PeterGhimself: Fixed, thanks

robot/rospackages/src/task_handler/scripts/task_handler_server.py
MartensCedric: nothing turns me more on then a bunch of checklists

MartensCedric:
Now that this PR breaks the frontend, it adds a lot of priority to @JonathanSolvesProblems PR, which is #201 because we would not want to write some front end code twice.

Accepting this PR will break the camera feature in master. I have a few propositions.

Ideally, we don't want to break stuff in master. There's different ways we can go about this PR.

1)  We keep this PR opened, even approved, make a PR of #201 once it's done and approved, **we merge #201 into this PR (#248) and then merge into master**. This guarantees master stability, but it may take longer before this feature is in master.

2) We merge this PR into master, @PeterGhimself you would need to write a checklist of work left to do after #201 is indeed merged and we **create an issue and add high priority to it**. This makes the branch go into master but breaks the whole camera stuff on the front-end level.

Imho 1) is superior because that means any testing/events/outreach/video won't be on master until #201 is done and that the changes needed after it are done, because the frontend feature will be broken in master. Although I would like to hear your suggestions @Davidster @PeterGhimself @NinjaWeeaboo @KoaTheKoala

PeterGhimself:
I think it makes sense to wait until we have a PR for #201 because of the braking front end changes. Perhaps issue #151 ought to be an epic split into backend + frontend issues.

robot/rospackages/src/task_handler/scripts/task_handler_server.py
MartensCedric: It's hot too see checkboxes :hot_pepper: :fire: :checkered_flag: :ballot_box_with_check:

PeterGhimself:
After discussing with @MartensCedric I will attempt to update the frontend regarding the breaking changes, but without entering the scope of #201. I will do this by adding a simple dropdown with the choice of the 3 camera port aliases for the rover as well as either `/dev/video0` - `/dev/video2`. This way while #201 is being worked on this PR can actually be merged into master without breaking the frontend and then local testing can be easily done on the branch for #201 

Perhaps I should add a text input field for custom input?

Davidster:
I would've agreed that option 1) is the best, though instead of merging 201 into this branch, why not just merge master into this branch once 201 is merged? Then you'll get changes from any other merges that might've happened at no additional cost.

But finally it seems that this branch can be updated such that it's totally compatible with 201? That sounds ideal to me.

image

NinjaWeeaboo:
> Imho 1) is superior because that means any testing/events/outreach/video won't be on master until #201 is done and that the changes needed after it are done, because the frontend feature will be broken in master. Although I would like to hear your suggestions @Davidster @PeterGhimself @NinjaWeeaboo @KoaTheKoala

I know I'm a bit late to the party and I don't know if you guys already made a decision, but I think we should go for option 1. I think it would make this whole process a lot cleaner in the long run, as opposed to merging this PR into master and then fixing the frontend with another PR.

robot/basestation/static/js/roslibjs/ros-helpers.js
```diff
@@ -318,6 +318,16 @@ function requestTask (
 }
 
 function checkTaskStatuses () {
+  if ($('#camera-local-mode-btn').length >= 1) {
+    let isLocal = getCookie('localCameras')
+    console.log('yoooooo')
```
MartensCedric: yoooooooooooo remove this

robot/rospackages/src/task_handler/scripts/start_ros_stream.sh
```diff
@@ -27,4 +27,22 @@ fi
...
+# Last test done on: 26/12/2019
+# +---------------------------------+--------------------------+
+# |resolution --> (width x height)  | max simultaneous streams |
+# |---------------------------------|--------------------------|
+# |240p  --> 352x240                | 3                        |
+# |360p  --> 480x360                | 2                        |
+# |480p  --> 858x480, SD            | 1                        |
+# |720p  --> 1280x720, HD           | 1                        |
+# |1080p --> 1920x1080, full HD     | 0                        |
+# +---------------------------------+--------------------------+
```
MartensCedric: Should be moved to the wiki (?)

PeterGhimself: I think that if we're at comp and we don't have internet it would be super useful to have this ready to read in the relevant place, in case something bad happens during comp. We should discuss more how to go about documenting our stuff in an organized manner. I think we should have most of it on the drive.

robot/basestation/static/js/roslibjs/ros-helpers.js
PeterGhimself: Done, thanks.

PR #263, opened 2020-01-01
Add ability to set emailer recipients by an external text file

### LINKED ISSUE #254: Make emailer read from list
Instead of having hardcoded emails in https://github.com/space-concordia-robotics/robotics-prototype/blob/master/robot/util/emailer/emailIPAddress.js

It should read a configuration file that is not on GH so we only need to update the file on the rover and not on GH

Davidster:
# Assignee Section

## Description
See PR name / issue description

closes #254 

# Reviewer Section

- [x] [General Integration Test](https://docs.google.com/document/d/1ug0CpA1cIzURP8DDFSvCt2CEJJSwJ6Ta6B1LG_hYk6I/edit) Performed Successfully

---

noice

### DIFF
--- a/robot/util/emailer/emailIPAddress.js
+++ b/robot/util/emailer/emailIPAddress.js
@@ -26,13 +26,13 @@ const getEmailsFromFile = async () => {
             .map(item => item.trim())
             .filter(item => item.length > 0);
         if (emails.length < 1) {
-            console.log(`Warning: no emails found in ${path.resolve(filePath)}. Falling back to default emails:`);
+            console.log(`Warning: no emails found in ${filePath}. Falling back to default emails:`);
             console.log(DEFAULT_EMAILS);
             emails = DEFAULT_EMAILS;
         }
     } catch (e) {
         if (e.code === 'ENOENT') {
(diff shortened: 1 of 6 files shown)


robot/util/emailer/emailIPAddress.js
```diff
@@ -1,28 +1,52 @@
+const fs = require("fs").promises;
+const path = require("path");
+const ip = require('ip');
+const nodemailer = require('nodemailer');
 
+const RETRY_DELAY = 3000;
+const MAX_TRIES = 40;
+const EMAILS_FILE_NAME = "emails.txt";
+const DEFAULT_EMAILS = [
```
PeterGhimself: Maybe add a comment here saying it's for in case there is not `emails.txt` file found?

robot/util/emailer/package-lock.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "lidar",
```
PeterGhimself: funny that this was auto generated, can we change it to something like `emailer`? or maybe `ip_emailer`?

PR #271, opened 2020-01-08
Custom module import statement test

### LINKED ISSUE #270: Proper import statement unit test
In [this](https://github.com/space-concordia-robotics/robotics-prototype/commit/2d7c76e7f89b7f6a14a655c3fd4fe24c92465979) commit we removed the motor_test unit test since we removed a class `motor.py` which was in `basestation` folder since we though it wasn't being used.

While it may have no functional use for the project, it is a good proof that our repo built correctly in that `setup.py develop` worked properly and we can import our own custom modules with the `robot.basestation.etc` format.

A new test should be made only to test that the import statements on our custom modules are working as they should after building the projects,

- [x] Add unit test to verify `python setup.py develop` suceeded during the build

PeterGhimself:
# Assignee Section

## Description
A simple test for verifying that `python setup.py develop` or `pip install -e` work correctly during project build phase and allow us to import our own custom modules.

closes #270

# Reviewer Section


Davidster:
Seems that this can be a bit useful but also a bit redundant. Wouldn't any unit test that uses the `robot.basestation.app` also fail if this one fails?

PeterGhimself:
@Davidster they are not explicitly testing for it, so when it fails for them they actually throw an error instead of a "failure". If other things crash because the robot module is not properly set up we won't have to guess when we see this test fail specifically for that. Tested on our conucourseplanner server for the negative cases btw.

Davidster:
👀👌

What do you mean by the negative cases?

PeterGhimself:
@Davidster by negative cases I mean negative testing vs positive. So for negative testing it means 
 feeding the test invalid inputs that would cause it to fail. In this case it's a lack of setting up the environment (or anything going wrong with that part of the build). Travis _does_ run `pip install -e .` which sets it up prior to running `pytest`. I'm considering maybe adding an alias [here](https://github.com/space-concordia-robotics/robotics-prototype/blob/master/robot/basestation/config/.bash_aliases#L31) to make sure `pytest` defaults to `pytest --continue-on-collection-errors` so that people our devs always run it with the proper each time without having to think about it. @MartensCedric thoughts?

tests/unit/import_test.py
```diff
@@ -0,0 +1,15 @@
...
+    ensure that build suceeded after running `python setup.py develop`
+    allowing for our custom modules to be imported anywhere in our repository
+    '''
+    from robot.basestation.app import app
+    assert not (app is None)
+
+def test_non_existant_import():
+    with pytest.raises(ImportError):
+        from robot.basestation.nonexistantmodule import nonexistantmodule
+        assert (app is None)
```
MartensCedric: `nonexistantmodule is None`

MartensCedric:
I tested locally, other than the issue I just pointed out its good to be accepted after it is fixed

tests/unit/import_test.py
PeterGhimself: oops :sweat_smile:

PR #319, opened 2020-02-22
Scp image transfer 192

badgerwild:
# Assignee Section

## Description
Created a script to send images using scp. The file name is passed to the script as a command line argument. In order for the script to work, ssh keys need to be setup on both the local and the host computers. This only has to be done once for each computer, before running the script.

Host address, host username, and remote file path all are updated in the config.py file so that this information does not need to be changed in the script itself.  

[If applicable, include a list of main points]
- [ ] Image transfers between computers

closes #[192]

# Reviewer Section

### DIFF
--- a/robot/rospackages/src/image_scp/config.py
+++ b/robot/rospackages/src/image_scp/config.py
@@ -1,0 +1,6 @@
+#enter host computer address
+host = "HOST_URL"
+#enter user name for host computer
+user = "USER NAME FOR HOST"
+#enter desired image path on host computer
+remote_path_1 = "PATH FOR IMAGE ON HOST COMPUTER"
--- a/robot/rospackages/src/image_scp/image_scp.py
+++ b/robot/rospackages/src/image_scp/image_scp.py
@@ -1,15 +1,21 @@
 from paramiko import SSHClient
 from scp import SCPClient, SCPException
 import argparse
-from congif import (host, user, remote_path_1, local)
+from config import (host, user, remote_path_1)
+"""
+Script takes in the image as a command line argument.
+SSH keys need to be configured on both remote and local machine in order for the script to work.
+Host name, user nmae, and remote path are all imported from a config file. Update the information there 
+so that this script is portable .
 
+"""
 class remote_connection:
     def __init__(self, host, user, remote_path, file_name):
         self.host = host
         self.user = user
         self.remote_path = remote_path
         self.file_name = file_name
-
+#initiate the connection to the host machine
     def connect(self):
         self.client = SSHClient()
         self.client.load_system_host_keys()
@@ -27,6 +33,7 @@ class remote_connection:
             raise error
 
 if __name__ == "__main__":
+    #takes the image as a command line argument
     parser = argparse.ArgumentParser(description="the file that gets sent")
     parser.add_argument("image_file", action="append")
     args = parser.parse_args()
@@ -36,16 +43,3 @@ if __name__ == "__main__":
     sender.connect()
     sender.send_image(sender.file_name)
     sender.disconnect()
-
-
-    print(args.image_file[0])
-
-
-#ssh = SSHClient()
-#ssh.load_system_host_keys()
-#ssh.connect(hostname="login.encs.concordia.ca", username='user')
-
-#scp = SCPClient(ssh.get_transport())
-
-#scp.put('/test2.png', recursive=True,
- #       remo te_path='remotepath    e)
--- a/robot/rospackages/src/image_scp/image_scp.py
+++ b/robot/rospackages/src/image_scp/image_scp.py
@@ -1,16 +1,11 @@
 from paramiko import SSHClient
 from scp import SCPClient
-import argparse
-from getpass import getpass
-
-
-parser = argparse.ArgumentParser(description="ssh login name")
-parser.add_argument(
-
+from config import (host, user, remote_path_1, local)
 ssh = SSHClient()
 ssh.load_system_host_keys()
-ssh.connect('j_klimo@login.encs.concordia.ca')
+ssh.connect(hostname=host, username=user)
 
 scp = SCPClient(ssh.get_transport())
 
-scp.put('test', recursive=True, remote_path='/home/j/j_klimo/Documents/images')
+scp.put(local, recursive=True,
+        remote_path=remote_path_1)
--- a/robot/rospackages/src/image_scp/image_scp.py
+++ b/robot/rospackages/src/image_scp/image_scp.py
@@ -1,6 +1,7 @@
 from paramiko import SSHClient
-from scp import SCPClient
-#from congif import (host, user, remote_path_1, local)
+from scp import SCPClient, SCPException
+import argparse
+from congif import (host, user, remote_path_1, local)
 
 class remote_connection:
     def __init__(self, host, user, remote_path, file_name):
@@ -19,7 +20,25 @@ class remote_connection:
         self.client.close()
         self.scp.close()
 
+    def send_image(self, image):
+        try:
+            self.scp.put(image, recursive=True, remote_path=self.remote_path)
+        except SCPException as error:
+            raise error
 
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="the file that gets sent")
+    parser.add_argument("image_file", action="append")
+    args = parser.parse_args()
+    to_transfer = args.image_file[0]
+
+    sender = remote_connection(host, user, remote_path_1, to_transfer)
+    sender.connect()
+    sender.send_image(sender.file_name)
+    sender.disconnect()
+
+
+    print(args.image_file[0])
 
 
 #ssh = SSHClient()
@@ -29,4 +48,4 @@ class remote_connection:
 #scp = SCPClient(ssh.get_transport())
 
 #scp.put('/test2.png', recursive=True,
- #       remote_path='remotepath    e)
+ #       remo te_path='remotepath    e)
--- a/robot/rospackages/src/image_scp/image_scp.py
+++ b/robot/rospackages/src/image_scp/image_scp.py
@@ -1,11 +1,32 @@
 from paramiko import SSHClient
 from scp import SCPClient
-from config import (host, user, remote_path_1, local)
-ssh = SSHClient()
-ssh.load_system_host_keys()
-ssh.connect(hostname=host, username=user)
+#from congif import (host, user, remote_path_1, local)
 
-scp = SCPClient(ssh.get_transport())
+class remote_connection:
+    def __init__(self, host, user, remote_path, file_name):
+        self.host = host
+        self.user = user
+        self.remote_path = remote_path
+        self.file_name = file_name
 
-scp.put(local, recursive=True,
-        remote_path=remote_path_1)
+    def connect(self):
+        self.client = SSHClient()
+        self.client.load_system_host_keys()
+        self.client.connect(hostname=self.host, username=self.user)
+        self.scp = SCPClient(self.client.get_transport())
+
+    def disconnect(self):
+        self.client.close()
+        self.scp.close()
+
+
+
+
+#ssh = SSHClient()
+#ssh.load_system_host_keys()
+#ssh.connect(hostname="login.encs.concordia.ca", username='user')
+
+#scp = SCPClient(ssh.get_transport())
+
+#scp.put('/test2.png', recursive=True,
+ #       remote_path='remotepath    e)
--- a/robot/rospackages/src/image_scp/image_scp.py
+++ b/robot/rospackages/src/image_scp/image_scp.py
@@ -1,0 +1,16 @@
+from paramiko import SSHClient
+from scp import SCPClient
+import argparse
+from getpass import getpass
+
+
+parser = argparse.ArgumentParser(description="ssh login name")
+parser.add_argument(
+
+ssh = SSHClient()
+ssh.load_system_host_keys()
+ssh.connect('j_klimo@login.encs.concordia.ca')
+
+scp = SCPClient(ssh.get_transport())
+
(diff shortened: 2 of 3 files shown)


robot/rospackages/src/image_scp/config.py
```diff
@@ -0,0 +1,6 @@
+#enter host computer address
+host = "HOST_URL"
+#enter user name for host computer
+user = "USER NAME FOR HOST"
+#enter desired image path on host computer
+remote_path_1 = "PATH FOR IMAGE ON HOST COMPUTER"
```
MartensCedric, line 6: So what happens if we want to dynamically send an image. I want to call this script lets say, everytime I take a new picture. This picture has a new name. Perhaps you want to just put a folder that needs to be synced. But then we will have issues if there's leftover files.

robot/rospackages/src/image_scp/image_scp.py
```diff
@@ -0,0 +1,45 @@
+from paramiko import SSHClient
+from scp import SCPClient, SCPException
+import argparse
+from config import (host, user, remote_path_1)
+"""
+Script takes in the image as a command line argument.
+SSH keys need to be configured on both remote and local machine in order for the script to work.
+Host name, user nmae, and remote path are all imported from a config file. Update the information there
```
MartensCedric, line 8: user nmae

MartensCedric:
Any progress?

badgerwild:
> Any progress?

Yes, I've been working on it this week. Was thinking of the best way to implement your changes. Been trying a few different ways, should have it resolved soon.

robot/rospackages/src/image_scp/image_scp.py
```diff
@@ -0,0 +1,78 @@
+import fnmatch
+import os
+from paramiko import SSHClient
+from scp import SCPClient, SCPException
+import argparse
+from config import (host, user, remote_path_1)
+"""
+Script takes in the image as a command line argument.
+SSH keys need to be configured on both remote and local machine in order for the script to work.
+Host name, user nmae, and remote path are all imported from a config file. Update the information
```
MartensCedric, line 10:
```suggestion
Host name, user name, and remote path are all imported from a config file. Update the information
```

robot/rospackages/src/image_scp/image_scp.py
```diff
@@ -0,0 +1,78 @@
...
+there so that this script is portable .
```
MartensCedric, line 11:
```suggestion
there so that this script is portable.
```

robot/rospackages/src/image_scp/image_scp.py
```diff
@@ -0,0 +1,78 @@
...
+        self.host = host
+        self.user = user
+        self.remote_path = remote_path
+        self.file_name = file_name
+#initiate the connection to the host machine
+    def connect(self):
+        self.client = SSHClient()
+        self.client.load_system_host_keys()
+        self.client.connect(hostname=self.host, username=self.user)
+        self.scp = SCPClient(self.client.get_transport())
```
MartensCedric, line 27:
```suggestion
    def connect(self):
        """initiate the connection to the host machine""""
        self.client = SSHClient()
        self.client.load_system_host_keys()
        self.client.connect(hostname=self.host, username=self.user)
        self.scp = SCPClient(self.client.get_transport())
```

robot/rospackages/src/image_scp/image_scp.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def send_batch_images(self, image_directory):
+        bulk_tranfer = [self.send_image(image) for image in image_directory]
+        print(len(bulk_transfer) + "Images have been transfered")
+
+
+if __name__ == "__main__":
+
+  #takes the image as a command line argument
+    parser = argparse.ArgumentParser(description="the file that gets sent")
```
MartensCedric, line 47:
```suggestion
    #takes the image as a command line argument
    parser = argparse.ArgumentParser(description="the file that gets sent")
```

robot/rospackages/src/image_scp/image_scp.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def send_image(self, image):
+        try:
+            self.scp.put(image, recursive=True, remote_path=self.remote_path)
+        except SCPException as error:
+            raise error
+
+    def send_batch_images(self, image_directory):
+        bulk_tranfer = [self.send_image(image) for image in image_directory]
+        print(len(bulk_transfer) + "Images have been transfered")
```
MartensCedric, line 41:
```suggestion
        print(len(bulk_transfer) + " images have been transfered")
```

robot/rospackages/src/image_scp/image_scp.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def disconnect(self):
+        self.client.close()
+        self.scp.close()
+
+    def send_image(self, image):
+        try:
+            self.scp.put(image, recursive=True, remote_path=self.remote_path)
+        except SCPException as error:
+            raise error
```
MartensCedric, line 37:
```suggestion
            print(error)
```

MartensCedric: You dont have a try catch when you send the image (outside of the function) by catching the error and rethrowing it, nothing useful happens.

MartensCedric: besides the program crashing

robot/rospackages/src/image_scp/image_scp.py
```diff
@@ -0,0 +1,78 @@
...
+  #takes the image as a command line argument
+    parser = argparse.ArgumentParser(description="the file that gets sent")
+    parser.add_argument("image_directory", action="append")
+    args = parser.parse_args()
+    to_transfer = args.image_directory[0]
+
+    image_files = []
+
+    #Takes in a directory name from which a config file and jpeg files are extracted
+    for name in os.listdir('test-file/'):
```
MartensCedric, line 55: is this `test-file` meant to be used in production or is this a mistake?

badgerwild: No, not meant for production. It's a mistake left over from testing. I'm  addressing your suggestions now.

badgerwild: @MartensCedric  Fixed this issue and debugged the rest of the code. It "Works on my machine" now.

NinjaWeeaboo:
Maybe this can be tested on the TX2 for those pro gamer bonus points

robot/rospackages/src/image_scp/image_scp.py
```diff
@@ -0,0 +1,79 @@
...
+from paramiko import SSHClient
+from scp import SCPClient, SCPException
+import argparse
+from config import (host, user, remote_path_1)
+"""
+Script takes in a directory name as a command line argument. This directory should contain
+a config.txt file and any images that need to be transered.
+SSH keys need to be configured on both remote and local machine in order for the script to work.
+Host name, user name, and remote path are all imported from config.txt file. Update the information
+there so that this script is portable.
```
MartensCedric, line 12:
```suggestion
Script takes in a directory name as a command line argument. This directory should contain
a config.txt file and any images that need to be transfered.
SSH keys need to be configured on both remote and local machines in order for the script to work.
Host name, user name, and remote path are all imported from config.txt file. Update the information
there so that this script is portable.
```

robot/rospackages/src/image_scp/config.txt
```diff
@@ -0,0 +1,6 @@
+#Use this as a template for entering user details and destination of files so script doesn't need to be changed.
+#Enter credentials/ path without quotation marks 
+Username replaceWithUse
```
PeterGhimself, line 3: you meant replaceWithUser? I would suggest to make it more clear like `basestationUserName`, though perhaps (since I recommend using a python file) it can be set automatically, by getting the `$USER` environement variable through python.

robot/rospackages/src/image_scp/config.txt
```diff
@@ -0,0 +1,6 @@
...
+Host relaceWithHostDomain
```
PeterGhimself, line 4: while host domain is probably still correct, it's a bit general. This is supposed to be the OBCs IP address right? Maybe we can just put `roverIP` instead?

robot/rospackages/src/image_scp/image_scp.py
```diff
@@ -0,0 +1,79 @@
+import fnmatch
+import os
+from paramiko import SSHClient
+from scp import SCPClient, SCPException
+import argparse
+from config import (host, user, remote_path_1)
+"""
+Script takes in a directory name as a command line argument. This directory should contain
```
PeterGhimself, line 8: Unforuntately I think this approach isn't the most efficient, it's probably best to make the script target a directory that is remote, the config file can stay local though.

robot/rospackages/src/image_scp/image_scp.py
```diff
@@ -0,0 +1,79 @@
...
+a config.txt file and any images that need to be transered.
+SSH keys need to be configured on both remote and local machine in order for the script to work.
+Host name, user name, and remote path are all imported from config.txt file. Update the information
+there so that this script is portable.
+
+"""
+
+class remote_connection:
```
PeterGhimself, line 16: Classes should be PascalCase, refer to the [TLDR](https://namingconvention.org/python/)

robot/rospackages/src/image_scp/config.txt
badgerwild: I added instructions for testing,  do they make sense?

PeterGhimself:
So I merged master in and added some changes after testing, there is however what I think is a bug/weird behavior.

To reproduce:

1. On host computer make sure you have `/home/[username]/Pictures/` folder with something in it (quickly: `touch yo`).
2. Run `./image_scp.py`, note that a new folder will be created in your home directory called `Picturez` with the files in there
3. Repeat step 2, but notice that it creates a new folder `Picturez` inside of the already existing `Picturez` folder

:thinking: 

Also perhaps we should rename this script? at first our intention was to use this for images but it is in fact a very general purpose script. Could be used to copy over log files, videos, etc. There is no constraint (and I don't think we should impose one) for it being images specifically.

Thoughts?

badgerwild:
@PeterGhimself 
I got the behaviour...I'll work on a fix for that.

Renaming the script to a general file transfer makes a lot of sense to me. I'll rename it to file_scp .

badgerwild:
> So I merged master in and added some changes after testing, there is however what I think is a bug/weird behavior.
> 
> To reproduce:
> 
>     1. On host computer make sure you have `/home/[username]/Pictures/` folder with something in it (quickly: `touch yo`).
> 
>     2. Run `./image_scp.py`, note that a new folder will be created in your home directory called `Picturez` with the files in there
> 
>     3. Repeat step 2, but notice that it creates a new folder `Picturez` inside of the already existing `Picturez` folder
> 
> 
> thinking
> 
> Also perhaps we should rename this script? at first our intention was to use this for images but it is in fact a very general purpose script. Could be used to copy over log files, videos, etc. There is no constraint (and I don't think we should impose one) for it being images specifically.
> 
> Thoughts?
@PeterGhimself 
I addressed the issue raised here about the unexpected behaviour for file creation.

MartensCedric:
Could you provide better steps for testing for I can get some people unfamiliar with the issue to test it. You should explain how to test it locally (what values to put in transferconfig), how to run the scripts, the arguments and the expected output. Then Ill test it and so should @RealWilliamWells and @AFknUsername

badgerwild:
@MartensCedric Updated the testing instructions.

MartensCedric:
@badgerwild I'm assuming I goofed somewhere 
image

Can you specify in the steps the exact ip that you used an all

image

MartensCedric:
Just for the future, when you put steps I would suggest to just copy put commands instead. Suppose you didn't edit "transferconfig.py" and someone not familiar with that specific file reviews this PR, he will have to find it. I found it easily because it was an edited file but if it wasn't other reviewers might have trouble finding it, and doing this small hassle for every other PR. That's why I would do steps like this. 

1. vim ~/Programming/robotics-prototype/robot/basestation/file_scp/transferconfig.py
2. Change rover ip to be "encs.concordia.ca"
3. Change rover username to be "(Your ENCS user name)"
...

badgerwild:
@MartensCedric  
The domain for concordia should be "login.encs.concordia.ca" Also, it looks like I goofed and didn't commit the updated config file, so it wouldn't have worked anyways. I'll update the steps following your recommendations .

ginywiny (commented):
Upon running with the new config, I ended up with the following issue. 
image

Has this ever happened to you @MartensCedric 

I have followed the new instructions, and set `rover_ip` to 'login.encs.concordia.ca', `rover_user` to my encs username 'm_lesk', I created a directory under my encs account called 'SpaceConcordia' and set `rover_image_dir` to '~/SpaceConcordia', lastly I set `payload_file_subdir` to 'TestFile', while I left `key_path` and `rsa_password` untouched.

badgerwild:
@ginywiny 
I'll try and have a look at that later today and see if I get issues.

AFknUsername:
I get the same error as @ginywiny when using the same settings
I also tried setting a custom ssh key to see if it would make a difference, and it does not work. I get an error :`There is a problem with the key, double check that your encryption password is correct` 
The key is not encrypted.

badgerwild:
> I get the same error as @ginywiny when using the same settings
> I also tried setting a custom ssh key to see if it would make a difference, and it does not work. I get an error :`There is a problem with the key, double check that your encryption password is correct`
> The key is not encrypted.

I'll try and figure out what is happening.

badgerwild:
@ginywiny @VimFreeman Following the stand up today, I changed the ssh key from rsa to ed25519 and that seemed to fix the problems we were having. The ed25519 also seems to produce a significantly faster runtime for the script when compared to the rsa key.

VimFreeman:
Closing since we do not need it anymore.

Issue #202, opened 2019-10-26
Add logging level filtering to log console


MartensCedric:
- [ ] Create a subscriber to `rosout` in order to get the logs from ROS
- [ ] Create a control to select desired logging levels (radio checkboxes)
- [ ] Filter the logs and display them based on the user's logging level

MartensCedric:
Considering adding the name of the node to the log

PR #322, opened 2020-02-25
[#321] Set the spammy logs to logDebug and add some logInfos

### LINKED ISSUE #321: Arm motor stutters when the page has been loaded for a while
Upon our early investigation (me and @MartensCedric ) it appears to be because the logs are clogged. Cedric recommend to not spam the arm commands in the log as much as we are doing right now.

MartensCedric:
# Assignee Section

## Description
The arm was constantly stuttering after playing it with for a while. The fix was to remove the spam from the log console, cluttering the page.

- Added minimum logging level. (See #202, where this feature should be expanded on)
- Put most of the spammy logs in debug level, which is not used by default
- Added some logs into logInfo and logDebug,  

closes #321 

# Reviewer Section

### DIFF
--- a/robot/basestation/static/js/roslibjs/ros-helpers.js
+++ b/robot/basestation/static/js/roslibjs/ros-helpers.js
@@ -24,7 +24,7 @@ const ROSFATAL = 16 // fatal/critical level
 
 
 // logs below this level will not be published or printed. Issue #202 will allow the user to set this value
-const MINIMUM_LOG_LEVEL = 2 
+const MINIMUM_LOG_LEVEL = ROSINFO 
 
 function initRosWeb () {
   ros = new ROSLIB.Ros({
--- a/robot/basestation/static/js/roslibjs/ros-helpers.js
+++ b/robot/basestation/static/js/roslibjs/ros-helpers.js
@@ -22,6 +22,10 @@ const ROSWARN = 4  // warning level
 const ROSERROR = 8 // error level
 const ROSFATAL = 16 // fatal/critical level
 
+
+// logs below this level will not be published or printed. Issue #202 will allow the user to set this value
+const MINIMUM_LOG_LEVEL = 2 
+
 function initRosWeb () {
   ros = new ROSLIB.Ros({
     url: 'ws://' + env.HOST_IP + ':9090'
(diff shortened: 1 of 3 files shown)


MartensCedric:
@vashmata can you review

robot/basestation/static/js/roslibjs/ros-helpers.js
```diff
@@ -641,24 +644,22 @@ function sendRequest (device, command, callback, timeout = REQUEST_TIMEOUT) {
...
 }
 
 /*
 returnsthe IP portion of the currntly set ROS_MASTER_URI
 */
 function getRoverIP (callback) {
-  console.log('roverIP: ' + env.ROS_MASTER_IP)
-  console.log('hostIP: ' + env.HOST_IP)
+  logDebug('roverIP: ' + env.ROS_MASTER_IP)
+  logDebug('hostIP: ' + env.HOST_IP)
```
vashmata: What's your reasoning for setting these as debug but not the ones on top like GUI colour theme?

MartensCedric: The GUI color theme is information that an operator can understand. This is debug information that would be useful to only someone in the software team

MartensCedric: Perhaps we should try to indicate when exactly each log should be used.

How I see it 

Info : Information that is operator readable (Operator can be mech or elec person, and may have no knowledge or understanding of the codebase).
Debug : Information useful for the software team if something is wrong. We should avoid putting debug on everything.
Warn : Warns in a situation that shouldn't happen or a behavior that's not normal but it isn't necessarily causing an issue.
Error : This is just not normal
Fatal : Fuck

robot/basestation/static/js/roslibjs/ros-helpers.js
```diff
@@ -22,6 +22,10 @@ const ROSWARN = 4  // warning level
 const ROSERROR = 8 // error level
 const ROSFATAL = 16 // fatal/critical level
 
+
+// logs below this level will not be published or printed. Issue #202 will allow the user to set this value
+const MINIMUM_LOG_LEVEL = 2
```
MartensCedric: Actually, I should use the constant ROSINFO here but yeah

robot/basestation/static/js/roslibjs/ros-helpers.js
vashmata: generally i agree but since the ip address stuff only happens once i think it's fine to leave it as info, it's not a spammy kind of log message

robot/basestation/static/js/roslibjs/ros-helpers.js
vashmata:
```suggestion
const MINIMUM_LOG_LEVEL = ROSINFO 
```

MartensCedric:
@vashmata ready for more review :sunglasses:

MartensCedric:
@PeterGhimself you too

MartensCedric:
Honestly this is a pretty important issue that we had trouble in the video, please review boys @PeterGhimself @vashmata :eye: :lips: :eye:

PR #323, opened 2020-02-25
[#310] Fix error handling in ping_acknowledgment

### LINKED ISSUE #310: Add error handling in `ping_acknowledgement_client` for no arguments
@vashmata was having issues when doing `rosrun ping_acknowledgement ping_acknowledgement_client`

because he did not provide any arguments. The default behavior should not be to crash but to still ping. "To your silence, I say 'suh dude'"

NinjaWeeaboo:
# Assignee Section

## Description
Improved error handling in the ping_acknowledgement scripts.

- [x] Fixed the preexisting error handling in the ping_acknowledgement scripts.
- [x] Added a case for when there is no argument inputted when running ping_acknowledgement_client.
- [ ] Added Fortnite dance ASCII art message to ping response, replacing the original text of 'suh dude'.

closes #310

## Testing

0. `roscore` in one terminal
1. `rosrun ping_acknowledgment ping_response_server.py` in another terminal
2. `rosrun ping_acknowledgement_client ping_acknowledgement_client' in another terminal, expect to see :eye: verbose output regarding ping statistics
3. `rosrun ping_acknowledgement_client ping_acknowledgement_client api', expect to see :eye: only the latency value (float)

# Reviewer Section


robot/rospackages/src/ping_acknowledgment/scripts/ping_response_server.py
```diff
@@ -5,11 +5,14 @@
 import rospy
 
 def handle_ping(req):
+    with open(os.environ['HOME'] + '/Programming/robotics-prototype/robot/rospackages/src/ping_acknowledgment/scripts/meme.txt', 'r') as meme:
```
MartensCedric: please don't use this small brain solution. Use a relative path : https://stackoverflow.com/questions/7165749/open-file-in-a-relative-location-in-python

NinjaWeeaboo: Fixed

MartensCedric:
After discussion, 

This package should return an (integer?) value of the ping in ms. Or maybe a bit more info. But it shouldnt return a string that needs to be parsed.

vashmata:
http://wiki.ros.org/rosping

NinjaWeeaboo:
I added a warning for when there's no string argument given and removed the Fortnite dance ascii art( sad day for all ). This package already returns the value of ping in ms along with the time the message was sent and received by the client. So what more info could it give? Also I think removing the string phrasing from this would defeat the purpose of having it as a custom package. Considering that there is already a ros package for pinging ros that returns the value of ping in ms.( See the link @vashmata posted here )

vashmata:
Is there actually a benefit to using the ping_acknowledgement service over just using rosping? Perhaps it would be simpler to just use rosping (except for the funny message aspect). If there's a difference, maybe it would be useful to implement both versions in the GUI so we would have more information.
@PeterGhimself

MartensCedric:
@NinjaWeeaboo From my comment earlier "After discussion,

This package should return an (integer?) value of the ping in ms. Or maybe a bit more info. But it shouldnt return a string that needs to be parsed."

MartensCedric:
The thing is we want to be able to take that value and put it in the GUI.

Right now if you want to get the value in an integer You have to select the second word from the third line and then cast to float lol.
image

MartensCedric:
@NinjaWeeaboo get fuked

NinjaWeeaboo:
OOF, I see what you mean now MB.

NinjaWeeaboo:
I decided to store the ping as an integer in a ROS parameter as a way of returning the ping from the package.

vashmata:
Given that parameters are more for settings, I think it would be more appropriate to publish it. If you want the last ping which was transmitted to be "held onto" by ROS you can set `latch=True`, [see here](http://wiki.ros.org/rospy/Overview/Publishers%20and%20Subscribers#rospy.Publisher_initialization) for more info.

MartensCedric:
> Given that parameters are more for settings, I think it would be more appropriate to publish it. If you want the last ping which was transmitted to be "held onto" by ROS you can set `latch=True`, [see here](http://wiki.ros.org/rospy/Overview/Publishers%20and%20Subscribers#rospy.Publisher_initialization) for more info.
> 
> Also, why not a float? I don't think you'll be saving that much bandwidth by using an int for a single variable.

true idk why I suggested an INT, that's kind of small brain. A float would likely be more appropriate

robot/rospackages/src/ping_acknowledgment/scripts/ping_response_client.py
```diff
@@ -47,3 +52,5 @@ def usage():
 
     diff = received - sent
     print("Latency: " + str(diff.total_seconds() * 1000) + " ms")
+    ping = rospy.Publisher('/ping_value', Float32, queue_size=10, latch=True)
+    ping.publish(Float32(diff.total_seconds() * 1000))
```
MartensCedric, line 56: So this is not quite what was intended. What I mean is the server should just return the amount directly. Because right now what you have to do is send the ping and then subscribe, we need to be able to do

`rosrun ping_ack ping_ack_client` and just get a number as a result. Rn we get a shitty string like this

```
Warning: no message argument, sending 'silence' to ping response server
usage: /home/cedric/Programming/robotics-prototype/robot/rospackages/src/ping_acknowledgment/scripts/ping_response_client.py [msg]
Pinging server with ping_acknowledgment service
2020-05-27T18:38:14-44
---
Response: To your 'silence', I say 'suh dude'
2020-05-27T18:38:14-44
Latency: 3.799 ms
```
It's completely useless and hard to parse, and to get the actual float value you would need to be subscribed to a topic. But even if you are subscribe to said topic, **you can't tell if you're the one who initiated the ping**

robot/rospackages/src/ping_acknowledgment/scripts/ping_response_client.py
```diff
@@ -3,54 +3,59 @@
...
+#
+# ping_response_client
+# 
+# Use without arguments to obtain detailed timestamps and latency data
+# Use option "api" to just receive the time it took in millisecond for the ping.
+#
 
-def ping_response_client(msg):
+
+def ping_server():
```
PeterGhimself, line 16: Why is this renamed to `ping_server`, isn't this the client script?

MartensCedric: to send a ping to the server

PR #345, opened 2020-03-14
[#316] Add more descriptive error message

### LINKED ISSUE #316: Microcontroller should respond with valid angle limits for motor if angle is out of bounds
Would have saved a lot of trouble for @vashmata.

vashmata:
# Assignee Section

## Description
When commands are sent to the arm mcu to move a joint out of bounds, the mcu will now respond with the angle limits, thereby alerting the operator what range they're allowed to be in.

### Instructions for testing this PR
This test will not involve the rover because it can be tricky to get it into a configuration in which the angles are out of bounds, unless I (Josh) am present. We could do that together once you've tested it locally though.
1) Take a spare teensy 3.6 and upload Arm.ino onto it.
2) Open up the Arduino serial console
3) Send a command which is out of bounds to the teensy, such as "motor 2 angle 400"
4) Expected response is an error message which alerts you of the angle limits for that particular joint. Motors 1 and 5 may not have any angle limits. In that case an error message will never be generated.

closes #316

# Reviewer Section

### DIFF
--- a/robot/rover/Arm/Arm.ino
+++ b/robot/rover/Arm/Arm.ino
@@ -439,7 +439,10 @@ void loop() {
 #if defined(DEVEL_MODE_1) || defined(DEVEL_MODE_2)
                   UART_PORT.print("ARM $E,Error: requested motor ");
                   UART_PORT.print(i + 1);
-                  UART_PORT.println(" angle is not within angle limits.");
+                  UART_PORT.print(" angle is not within angle limits: ");
+                  UART_PORT.print(motorArray[i] -> minJointAngle);
+                  UART_PORT.print(" and ");
+                  UART_PORT.println(motorArray[i] -> maxJointAngle);
 #elif defined(DEBUG_MODE) || defined(USER_MODE)
                   // this is SUPER DUPER GROSS
                   int tempVal = i + 1;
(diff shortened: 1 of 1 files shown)


MartensCedric:
Not requiring GIT given that I was there when the bug happened and that the changes are simply about logging

vashmata:
> I'm not going to roast you on the fact that you're using a space between the arrow which is very meme to me.

Personally i think it's just more readable because it's a 2 character long operator.

robot/rover/Arm/Arm.ino
```diff
@@ -439,7 +439,10 @@ void loop() {
 #if defined(DEVEL_MODE_1) || defined(DEVEL_MODE_2)
                   UART_PORT.print("ARM $E,Error: requested motor ");
                   UART_PORT.print(i + 1);
-                  UART_PORT.println(" angle is not within angle limits.");
+                  UART_PORT.print(" angle is not within angle limits: ");
+                  UART_PORT.print(motorArray[i] -> minJointAngle);
+                  UART_PORT.print(" and ");
+                  UART_PORT.println(motorArray[i] -> maxJointAngle);
```
PeterGhimself, line 445: Please do not separate -> or . accessor operators in C++ by spaces

PeterGhimself:
```suggestion
                  UART_PORT.print(" angle is not within angle limits: ");
                  UART_PORT.print(motorArray[i]->minJointAngle);
                  UART_PORT.print(" and ");
                  UART_PORT.println(motorArray[i]->maxJointAngle);
```

PR #353, opened 2020-03-21
Stop front-end from yelling at you for 999.9 temp

### LINKED ISSUE #317: Show N/A as temperature when they temperature is not available instead of 999 degrees
Putting 999 degrees triggers an alert every time we open the page. 

The text written should be N/A and the alert should not pop up

VimFreeman:
# Assignee Section

## Description
Implement a check to replace the element with "N/A" if the temp is equal to 999.9
I don't like this fix, but am not allowed to change PDS code so it is what it is.

### Steps for Testing
1) You need the rover
2) You need the rover

closes #317

# Reviewer Section


MartensCedric:
I lol'ed at the title

AFknUsername:
@MartensCedric it checks for 999.9 i just confirmed with PDS code. I updated PR details. 
For the ==, the data string from PDS is parsed and cast into float() so I think it's safe. 
The PDS sends 999.9 as a default value.

vashmata:
I suggest that instead of checking for 999, maybe pick a safe range like above 100V and below 0V as these values are way outside the acceptable range.

MartensCedric:
These are temperatures josh but the idea is great. We can expect the values
to be between -100 and 500 C lol?

vashmata:
Let's just use +/-200 for both temps and voltages

AFknUsername:
I don't see how this is relevant. The default error value is 999.9 from the PDS code, so I'm turning that value into N/A for the front-end. Changing the value to N/A if its outside a range would do the same, but would use arbitrary values (i.e. +/-200)

robot/rospackages/src/mcu_control/scripts/PdsNode.py
```diff
@@ -205,9 +206,12 @@ def publish_pds_data(message):
...
             elif i < 10:
-                temp.x = float(dataPDS[7])
-                temp.y = float(dataPDS[8])
-                temp.z = float(dataPDS[9])
+                if i == 7:
+                    temp.x = checkTempErrorCase(float(dataPDS[i]))
+                if i == 8:
+                    temp.y = checkTempErrorCase(float(dataPDS[i]))
+                if i == 9:
+                    temp.z = checkTempErrorCase(float(dataPDS[i]))
```
MartensCedric, line 214: why not do this ?
```suggestion
                temp.x = checkTempErrorCase(float(dataPDS[7]))
                temp.y = checkTempErrorCase(float(dataPDS[8]))
                temp.z = checkTempErrorCase(float(dataPDS[9]))
```

robot/rospackages/src/mcu_control/scripts/PdsNode.py
```diff
@@ -241,6 +245,12 @@ def publish_pds_data(message):
         return
     return
 
+def checkTempErrorCase(x):
```
MartensCedric, line 248: You should maybe document this function to explain why it needs to be there. So some guy in the future that didnt know about the 999.9 meme popup can understand

robot/rospackages/src/mcu_control/scripts/PdsNode.py
MartensCedric: I guess its because previously it was ran 3 times? (Man that for loop is garbage with that magic `i`)

AFknUsername: I honestly don't know why I checked each case individually. 
Also lol yes that loop is great. Prob the person who wrote the PDS code also wrote this?

MartensCedric: image


Use `git blame` @AFknUsername

AFknUsername: @vashmata 👀 👀

robot/rospackages/src/mcu_control/scripts/PdsNode.py
```diff
@@ -205,9 +206,9 @@ def publish_pds_data(message):
             elif i < 7:
                 current.effort.append(float(dataPDS[i]))
             elif i < 10:
-                temp.x = float(dataPDS[7])
-                temp.y = float(dataPDS[8])
-                temp.z = float(dataPDS[9])
+                    temp.x = checkTempErrorCase(float(dataPDS[i]))
+                    temp.y = checkTempErrorCase(float(dataPDS[i]))
+                    temp.z = checkTempErrorCase(float(dataPDS[i]))
             else:
```
MartensCedric, line 212:
```suggestion
                temp.x = checkTempErrorCase(float(dataPDS[i]))
                temp.y = checkTempErrorCase(float(dataPDS[i]))
                temp.z = checkTempErrorCase(float(dataPDS[i]))
            else:
```

robot/rospackages/src/mcu_control/scripts/PdsNode.py
```diff
@@ -241,6 +242,15 @@ def publish_pds_data(message):
         return
     return
 
+# This function checks if the temp values from serial are equal to 999.9 (deafult error state) and replaces them with 'N/A' if they are.
+# It is necessary because if the temperature is at the default error state, the front end will scream at you with a warning. 
+# Changing it to 'N/A' prevents being assaulted by UI.
+def checkTempErrorCase(x):
```
MartensCedric, line 248:
```suggestion
# This function checks if the temp values from serial are equal to 999.9 (default error state) and replaces them with 'N/A' if they are.
# It is necessary because if the temperature is at the default error state, the front end will scream at you with a warning with a horrendous Windows sound. 
# Changing it to 'N/A' prevents being assaulted by UI.
def checkTempErrorCase(x):
```

MartensCedric:
I approved but I opened #374 since a lot of stuff is garbage in the PDS node.

PeterGhimself:
We're gonna have to wait until we can test with the rover for merging this one, but the changes seem fine.

PeterGhimself (dismissed):
Tested with the rover, imo needs a bit more work:

1. Since you are publishing a string type from the PDS node of message type "ThermistorTemp", but it is declared as a float64, there is a conflict which causes an error:

```
  File "/home/odroid/Programming/robotics-prototype/robot/rospackages/devel/lib/python3/dist-packages/mcu_control/msg/_ThermistorTemps.py", line 61, in serialize
    buff.write(_get_struct_3d().pack(_x.therm1, _x.therm2, _x.therm3))
struct.error: required argument is not a float

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/rospy/topics.py", line 882, in publish
    self.impl.publish(data)
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/rospy/topics.py", line 1066, in publish
    serialize_message(b, self.seq, message)
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/rospy/msg.py", line 152, in serialize_message
    msg.serialize(b)
  File "/home/odroid/Programming/robotics-prototype/robot/rospackages/devel/lib/python3/dist-packages/mcu_control/msg/_ThermistorTemps.py", line 62, in serialize
    except struct.error as se: self._check_types(struct.error("%s: '%s' when writing '%s'" % (type(se), str(se), str(locals().get('_x', self)))))
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/genpy/message.py", line 336, in _check_types
    raise SerializationError(str(exc))
genpy.message.SerializationError: <class 'struct.error'>: 'required argument is not a float' when writing 'therm1: "N/A"
therm2: "N/A"
therm3: "N/A"'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/odroid/Programming/robotics-prototype/robot/rospackages/src/mcu_control/scripts/PdsNode.py", line 118, in publish_pds_data
    tempPub.publish(temp)
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/rospy/topics.py", line 886, in publish
    raise ROSSerializationException(str(e))
rospy.exceptions.ROSSerializationException: <class 'struct.error'>: 'required argument is not a float' when writing 'therm1: "N/A"
```

one workaround/solution would be changing the message type to `string, string string`, rather than `float64, float64, float64`, which I tested and it works. But this makes me think perhaps this should rather be purely handled by the front end? I.e. the front end sees the specific value 999.9 and handles the edge case by changing the corresponding element content to `N/A`.

2. Lets suppose we set the msg type of ThermistorTemps to `string, string, string`, it will work now, except in the GUI it will show `NaN` for those values (I believe the cause would be [these](https://github.com/space-concordia-robotics/robotics-prototype/blob/na-temp-warning-317/robot/basestation/static/js/roslibjs/ros-helpers.js#L247) lines). If we go about it this way, then for completeness I would say it's probably best to handle let the frontend handle it so it shows `N/A` rather than `NaN`.

MartensCedric (changes requested):
Please implement the previously requested change

Issue #312, opened 2020-02-17
Investigate ROS package installation: apt install, git clone, or both?


vashmata:
@MartensCedric claims that certain packages, at the very least cv_camera, does not build correctly if it is simply git cloned into rospackages. Furthermore, in the quick setup script which can be used to install ROS and set up a new linux machine for our repo, it calls apt install for certain ROS packages. That said, the ideal scenario is one in which all external packages are linked to our repo and reside in rospackages.

Someone must investigate whether all of our packages can work if we simply git clone them into our repo, or if some packages require installation via apt, or if some packages require both installation methods to be used.

MartensCedric:
According to ROS best practices : 

https://www.ros.org/reps/rep-0128.html

git clone > catkin_ws

or sudo apt get install is fine. 

What we are doing is pretty small brain since our rospackages acts as a catkin_ws


With my job, I'm working with multiple ROS project and the catkin_ws approach is more convenient

MartensCedric:
Also your cmake targets should go in devel not where we have them, anwyay. That's according to the "recommended layout"

MartensCedric:
Removing the good first issue tag because with my suggestion it involves a bit of work to actually make the changes

MartensCedric:
Perhaps we can put our dependencies in `package.xml` of task_handler so then we can just run rosdep and have cvcamera and all being downloaded @PeterGhimself

PR #356, opened 2020-04-05
Setup TX2 initialization

### LINKED ISSUE #275: Setup TX2 initialization
We are switching to the Jetson TX2 as our new OBC away from the Odroid XU4.

We know that through @KoaTheKoala 's efforts we have ubuntu 16.04 and ros kinetic installed on our TX2. Now we must setup the repo and systemd service to ensure that on startup everything will behave the same way as before. There are probably going to be a number of breaking changes that need to be dealt with.

RealWilliamWells:
# Assignee Section

## Description
Made some code changes and additions to ensure that ROS works properly on the TX2.

- Modified the systemd services and startup scripts that run on the OBC in order for them to work with the user directories of the TX2.
- Created a setup script( not tested very much, likely has issues ), to keep track of installed software and to make setting up OBC go more smoothly in the future.

### Steps for Testing
1) Follow the client section of this [guide](https://docs.google.com/document/d/1N7UnQm65c9YOszqaOxbKguqmpEOwiNORGUgtopzSDnI/edit) for setting up OpenVPN, in order to be able to connect to the Jetson.( Feel free to leave comments in the document to let me know if my instructions suck )
2) Run `ping nividia` to ensure that you are able to communicate with the Jetson.
3) Ssh into the Jetson by using `ssh nvidia@nvidia` and the password `nvidia`( this is also the sudo password ). Add your computer's ip and username to its /etc/hosts file.
3) Run `rosgui` and `startgui`.
4) Run the ros `ping_acknowledgement_client` node to test your computer's connection to ros master on the Jetson.
5) Run `rostopic list` and `rosnode list` on the Jetson to view currently subscribed rostopics and running rosnodes.
5) Open the gui and test run anything else and manually run any other ros packages that you can think of.
6) Run the setup script in a VM or on another user account and ensure it sets everything up properly. Make the sure the user you run it on is named "Nvidia"! ( VM recommended this will install systemd services! ).

**Note:**
For some odd reason, the ros nodes use the Jetson's computer name of 'jetson-1422719019290' as its hostname instead of its username, when using the VPN server and ran from the systemd service. Roscore still uses the hostname 'nvidia', which is why you have to assign its ip to both of these hostnames. This behaviour didn't seem to occur when connecting over ethernet. 

closes #275

# Reviewer Section


robot/rover/RoverSetup.sh
```diff
@@ -0,0 +1,109 @@
...
+# Script which setups the Space Concordia Robotics Software team's OBC.
+
+APPEND_TO_BASH="
+
+export ROS_MASTER_URI=http://localhost:11311
+export ROS_HOSTNAME=$USER
+
+. ~/Programming/robotics-prototype/robot/rospackages/devel/setup.bash
+. ~/Programming/robotics-prototype/venv/bin/activate
+source ~/Programming/robotics-prototype/robot/basestation/config/.bash_aliases
```
MartensCedric, line 11: Shouldn't you use $REPO for this?

robot/rover/RoverSetup.sh
```diff
@@ -0,0 +1,109 @@
...
+sudo ufw allow ssh
+
+# Setup venv
+cd $REPO
+python3.6 -m venv venv
+source venv/bin/activate
+
+# Install Requirements
+pip install -U pip
+pip install pyserial
```
MartensCedric, line 45: is there a reason why pip is installed here and not ran through pip -r requirements? 

Also you don't include a version # which could be potentially bad

robot/util/emailer/emailIPAddress.js
```diff
@@ -9,7 +9,8 @@ const EMAILS_FILE_NAME = 'emails.txt'
 // fallback to default emails if there's and issue with the emails file
 const DEFAULT_EMAILS = [
     'davidhuculak5@gmail.com',
-    'petergranitski@gmail.com'
+    'petergranitski@gmail.com',
+    'william.wells@spaceconcordia.ca'
```
MartensCedric, line 13: lol pls add me

MartensCedric: Just so you know, read this code to see how to place the email config file so that more members can be notified

robot/util/rosRoverStart/rosRoverStart.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-HOME="/home/odroid"
+HOME="/home/nvidia"
```
MartensCedric, line 2: why is it sometimes $USER sometimes nvidia

robot/rover/systemd/ros-rover-start.service
```diff
@@ -4,11 +4,11 @@ After=ip-emailer.service
...
 
 [Service]
-#User=odroid
-# For some unkown reason this service stopped working when running as user odroid
+#User=nvidia
+# For some unkown reason this service stopped working when running as user nvidia
 User=root
 Type=simple
-ExecStart=/home/odroid/Programming/robotics-prototype/robot/util/rosRoverStart/runRosRoverStart.sh
+ExecStart=/home/nvidia/Programming/robotics-prototype/robot/util/rosRoverStart/runRosRoverStart.sh
```
MartensCedric, line 11: Shouldn't you be using $USER

robot/util/rosRoverStart/rosRoverStart.sh
MartensCedric: is this related to what you put in the description?

NinjaWeeaboo: There are apparently issues with using $USER when running it as a systemd service. I'll try to changing it back to see if I can get to work tho

NinjaWeeaboo: Also it isn't related to what I put in the description. The description refers to some weird issue with connecting to ROS over a vpn server which has a workaround.

robot/rover/RoverSetup.sh
NinjaWeeaboo: Not all of the python libraries are required, so there's no point to create a requirements.txt file just for one library. I included the version # thanks

robot/rover/RoverSetup.sh
NinjaWeeaboo: Fixed

robot/util/emailer/emailIPAddress.js
NinjaWeeaboo: Done

robot/util/rosRoverStart/rosRoverStart.sh
NinjaWeeaboo: The systemd service doesn't work if $USER is used anywhere because the service has to run under root for some reason. I haven't find a way to fix this yet

robot/rover/systemd/ros-rover-start.service
NinjaWeeaboo: Won't work cause the systemd service has to run under root for some reason

NinjaWeeaboo:
I think Travis failed to download one of the requirements on the first test and failed lmao

MartensCedric:
So how can we test this? You provided some steps but is the jetson always on? If not, can we coordinate a testing time? @KoaTheKoala did you ever test this since I know you did some jetson work

NinjaWeeaboo:
@MartensCedric We would prob have to coordinate a testing time, unless if @KoaTheKoala runs the jetson 24/7. The setup script I made can be tested in VM to ensure that there's no errors, although that doesn't guarantee that it would setup every properly on the Jetson.

PeterGhimself:
I would also suggest we should include some teensys into the equation that are running the code for rover/arm/science modules, to at least verify if the comms between the MCUs/OBCs still function.
With this you can try to start listeners in the GUI and see if sending commands "works". Without the real deal, we won't see anything budge, but I think the internal LEDs are programmed to blink accordingly (I can re-verify this claim later today).
As well as hypothetically maybe some kind of loads could be attached to represent wheels, maybe even just smaller but similar type motors (that way we can easily set them up to see if they are rotating correctly, more or less).

robot/rover/systemd/ros-rover-start.service
PeterGhimself: Thing is, you can change the service to run as non-root but then certain processes later in the pipeline will fail and need to be worked around. Some more digging can be done as to where specifically it messes up, but seeing as we're not going to be switching OBC super often I'd say it's fine to leave this as it is for now. Maybe we can open an issue to investigate this for after this PR is merged.

MartensCedric:
Given that its a pain for anyone remote to test, maybe we could devise a test for Koa to perform, upon success of this test we could merge the PR

robot/rover/RoverSetup.sh
```diff
@@ -0,0 +1,110 @@
...
+elif [$ROS_VERSION != "kinetic"]
+then
+    echo "A different ROS installation has been found... Please uninstall and rerun the script."
+    exit 1
+
+fi
+
+
+# Install camera stuff, these are not ros package dependecies and not installed with rosdep
+sudo apt-get install ros-kinetic-cv-camera ros-kinetic-web-video-server -y
```
PeterGhimself, line 76: There may be a better way, one that will allow us to remove this line specifically. Not a change request, but just a reminder that we may remove this line based off the outcome of #312

NinjaWeeaboo: Alrighty I will note this, encase there's is a pr that removes this line

robot/rover/RoverSetup.sh
```diff
@@ -0,0 +1,110 @@
...
+sudo ufw allow ssh
+
+# Setup venv
+cd $REPO
+python3.6 -m venv venv
+source venv/bin/activate
+
+# Install Requirements
+pip install -U pip
+pip install pyserial==3.4 rospkg==1.1.7
```
PeterGhimself, line 45: since these are included in `requirements.txt` perhaps we should just use that? i.e. `pip install -r requirements.txt -r requirements-dev.txt` for completeness

robot/rover/RoverSetup.sh
```diff
@@ -0,0 +1,110 @@
...
+    echo "You do not have ROS installed, installing..."
+
+    wget https://raw.githubusercontent.com/ROBOTIS-GIT/robotis_tools/master/install_ros_kinetic.sh
+    chmod 755 ./install_ros_kinetic.sh
+    yes "" | bash ./install_ros_kinetic.sh
+    rm ./install_ros_kinetic.sh
+
+	source ~/.bashrc
+
+    sudo apt install ros-kinetic-rosbridge-suite -y
```
PeterGhimself, line 64: This script needs to be run as sudo? If so maybe we can add a previous check to see if `whoami` returns `root`. If not, exit the script with a message saying it needs to be run with sudo.

robot/rover/RoverSetup.sh
```diff
@@ -0,0 +1,110 @@
...
+sudo systemctl enable config-ethernet.service
+sudo systemctl enable ip-emailer.service
+sudo systemctl enable ros-rover-start.service
+
+# Setup ethernet and emailer service scripts
+cd $REPO/robot/util
+sudo cp configEthernet/runConfigEthernet.sh /usr/bin/runConfigEthernet.sh
+sudo cp emailer/runEmailer.sh /usr/bin/runEmailer.sh
+cd $REPO/robot/util/configEthernet && bash synchConfigEthernet.sh
+cd $REPO/robot/util/emailer/ && bash syncEmailer.sh
```
PeterGhimself, line 106: I'd be surprised if the node emailer script worked out of the box. Gonna make sure to test this.

robot/rover/systemd/ros-rover-start.service
PeterGhimself: I can try testing this out again for some investigation :eyes: :wine_glass:

robot/rover/RoverSetup.sh
```diff
@@ -0,0 +1,36 @@
...
+#install prereqs
+sudo add-apt-repository ppa:deadsnakes/ppa -y
+sudo apt update -y
+sudo apt install nodejs npm openssh-server -y
+sudo ufw allow ssh
+
+# Setup venv
+cd $REPO
+
+# Setup systemd services
```
MartensCedric, line 18: This is useless

MartensCedric (changes requested):

robot/util/rosRoverStart/rosRoverStart.sh
```diff
@@ -21,4 +21,4 @@ source "$REPO_HOME/venv/bin/activate"
 python3 "$REPO_HOME/setup.py" develop
 
 # source primary catkin_ws setup bash script and execute one launch script to rule them all
-source $OPT_KINETIC_SETUP && source $ROS_PACKAGES_SETUP && roslaunch $ROSLAUNCH_FILE
+source OPT_MELODIC_SETUP && source $ROS_PACKAGES_SETUP && roslaunch $ROSLAUNCH_FILE
```
MartensCedric, line 24: dollar sign missing

MartensCedric (changes requested):

MartensCedric:
We had trouble compiling web_video_server until this was done : https://github.com/RobotWebTools/web_video_server/issues/113

Then two problems with the services

1. With ros-rover-start, the email doesn't send because it can't find "node" despite node working just fine in the terminal (but it doesnt work from the startup service)

2. config ethernet also crashes but this is probably just because there's nothing in the ethernet port

MartensCedric (changes requested):
Before merging this, the exact steps of the non-automated procedure should be written on a README or a wiki page. 

Something that are not quite well documented (they are documented somewhere but not in a central place)

- install npm
- services
- udev
- webvideoserver edit

PR #364, opened 2020-05-14
Fix commit message hook 363

### LINKED ISSUE #363: Commit message hook using incorrect number from branch name.
In some situations, the commit message hook that automatically puts the issue number in commit messages, can use another number that isn't the issue number from the branch name. 

Ex) Using the '2' from tx2 instead of the issue number '275' for the branch 'setup-tx2-initialization-275'. Causing the commit to be wrongly associated with issue 2 instead of issue 275. This has to be manually corrected with `commit --amend`.

NinjaWeeaboo:
# Assignee Section

## Description
Fixed the issue with the git commit message hook applying the wrong number from the branch name to the commit message. I didn't bother with making separate unit tests for this hook, as all of the input/output functions used are already tested.

- [x] Slightly refactored branch name verification hook to avoid reusing code.
- [x] Added new Python util function for writing text in files at a given location, used for applying the issue number to the commit message in '.git/COMMIT_EDITMSG'.
- [x] Edited readme with new name of commit hook.
- [x] Add new test case for `get_issue_num` function to ensure that it can properly get the issue number from 'setup-tx2-initialization-275'( the branch where this issue was discovered ).

### Steps for Testing
1) After checking out `fix-commit-message-hook-363`, run `cp commit_message_hook.py .git/hooks/prepare-commit-msg` to setup the new commit hook.
2) While still on 'fix-commit-message-hook-363', temporarily change the name of the branch to one that would fail before the changes in this PR, like `test-123-456` using `git branch -m test-123-456`.
3) Create a dummy file, add it to the staging area and commit it without pushing it.
4) Use `git log` to ensure that the number at of the end of the branch name, was put at the beginning of the commit message.
5) Run `git reset HEAD^` to undo the last commit, the dummy file should be automatically deleted.
6) Change the name of the branch back to the original with `git branch -m fix-commit-message-hook-363`
7) Run `cp commit-message-hook.sh .git/hooks/prepare-commit-msg` on master because the new hook will only work on this branch.

Running this on other branches will not work until if you bring the changes to `utils.py` and `branch_name_verification_hook.py` to those branches. I tested this on 'setup-tx2-initialization-275' and it does fix the issue that the old hook had with that branch name.

closes #363

# Reviewer Section


NinjaWeeaboo:
Wait I somehow merged a random pull request lol

commit_message_hook.py
```diff
@@ -0,0 +1,23 @@
...
+import re
+import sys
+
+def add_commit_issue_num(branch_name, error_msg):
+    """
+    Add commit issue number to commit message, return error if there is no issue number found in branch name
+    """
+    issue_num = get_issue_num(branch_name)
+
+    if not get_issue_num(branch_name) is None:
```
MartensCedric, line 14:
```suggestion
    if get_issue_num(branch_name) is not None:
```

More readable imho

robot/util/utils.py
```diff
@@ -36,3 +36,8 @@ def get_pid(keyword):
 
     return -1
 
+def write_to_file(filename, text, locationx, locationy):
```
MartensCedric, line 39: You're writing a high level function that is supposed to be abstract like "Write To File" but it takes arguments about the small details. I would try not to pass locationx and locationy, why do you need that? Did you just want to append?

branch_name_verification_hook.py
```diff
@@ -5,8 +5,12 @@
 
 excluded_branches = {'master', 'develop', 'staging', 'test'}
 
-output, error = run_shell('git', 'symbolic-ref --short HEAD', False)
-branch_name = output.decode('utf8')[:-1]
+wiki_url = 'https://github.com/space-concordia-robotics/robotics-prototype/wiki/Git-Workflow-and-Conventions'
+error_msg = 'Branch name is not named properly, please see wiki for formating: ' + wiki_url
```
MartensCedric, line 9: I prefered having the error message inside that function since its very odd that you need to pass an error message as an argument to a function that its role is to check if a branch is good or not. Also, it seems like you don't use it anywhere else.

commit_message_hook.py
NinjaWeeaboo: fixed

robot/util/utils.py
NinjaWeeaboo: Good point, removed arguments `locationx` and `locationy` and renamed function to `append_to_file`.

branch_name_verification_hook.py
NinjaWeeaboo: I did this in order to avoid recreating the `wiki_url` and `error_msg` variables in the commit message hook. So I did actually reuse it in the commit message hook. I guess it's not that big of a deal tho, so I'll just change that.

NinjaWeeaboo:
Nice, thanks man I never thought of doing that for testing

Issue #362, opened 2020-05-09
Automate Teensyduino library installs with CMake


MartensCedric:
We should have a section in the README explaining how to setup teensyduino as well as which libraries must be downloaded. Note that we should think how we would automate this.

PeterGhimself:
Isn't that explained [here](https://github.com/space-concordia-robotics/robotics-prototype/tree/master/robot/rover#teensy)?

MartensCedric:
> Isn't that explained [here](https://github.com/space-concordia-robotics/robotics-prototype/tree/master/robot/rover#teensy)?

Not quite,

When you first run the project it won't compile. You will be missing libraries. To get them, you have to search your error, find what libraries is missing, download it, rinse and repeat 10 times. The list of used libraries could be documented and a link to the document you points should be in the wiki or README if we want people to see it.

PeterGhimself:
Maybe we should just have like a `requirements.txt` sort of file to contain every dependency line by line and their version. Perhaps that could allow for some easier custom automation, if there is no existing framework to help automate. Btw, the libraries your refering to need to be downloaded through arduino IDE?

MartensCedric:
>  Btw, the libraries your refering to need to be downloaded through arduino IDE?

Yup, since its compiled with Arduino afaik, but @steamypotato had some scripty ideas

vashmata:
Arduino libraries can be downloaded from github and dropped into the libraries folder. That said, I haven't tested this all that much in linux, we will need someone to uninstall the libraries and try it out. It's possible that not all our libraries will "like" this approach.

vashmata:
As it is I got the impression that some of the libraries cedric is referring to don't play nicely, i remember my laptop being the only one capable of downloading code to the rover teensy at a few points

PeterGhimself:
> i remember my laptop being the only one capable of downloading code to the rover teensy at a few points

Yeah at URC 2019 that was the case I remember, but then I fixed the issue for my laptop. It's been a while so I'd have to dig a bit to see how I fixed it.

MartensCedric:
These are the libraries I have installed, the project probably needs less than that. The assignee of this task should determine which ones are actually needed.

![image](https://user-images.githubusercontent.com/18620025/82744915-b7814880-9d4c-11ea-9fbe-2d36ba70f63b.png)

[...]

PR #372, opened 2020-05-26
Cmake integration 204

### LINKED ISSUE #204: Use CMake to build the C++
CMake should be configured to build with the same compiler as the Arduino IDE.

steamypotato:
# Assignee Section

## Description
Uses CMake to build the Teensyduino projects. The design is inspired from unit tests, so it doesn't interact with the user, everything that is created can easily be instantly deleted. It uses out-of-source builds, to keep the folders clean. 
Lastly, quite a few things are subject to change whenever the eventual refactoring issue for the C++ code is done. An example of this is that there are quite a few non-essential projects in the workspace that are purposefully not built, so all the `continue` statements in the code are just temporary.
### Overview of design
The main CMakeLists.txt file resides in a root directory, which will be `robot/rover` in this case. It finds the Arduino/Teensy sources files on the computer, and compiles all the necessary source files into a few libraries. It recursively finds all the Arduino sketches from the root directory, and generates a `CMakeLists.txt` file for each sketch and runs it, using the `CMakeTemplate.txt.in` file, where the template take in a project name and used libraries. 
 Each sub-directory will contain a `CMakeLists.txt` file a `build` folder, with `bin` and `lib` folders in `build`, respectively containing the final `.elf` file for that specific Teensyduino project and a library created from the source codes in that sub-directory.    

### Steps for Testing

0. Make sure to have the arduino IDE + teensyduino installed (if not see steps below), as well these libraries throught the library manager. Go to `sketch -> include libraries -> manage libraries...` to open the Library manager and install the following libraries:

- - Adafruit Motor Shield V2 Library -> 1.0.9
- - ArduinoBlue -> 3.0.1
- - LSM303 by Polulu -> 3.0.1
- - SparkFun I2C GPS Reading and Control -> 1.0.2 (newest version 1.0.3 doesn't compile with teensy compiler)

1) Open `~/.bashrc`
2) Append `export ARDUINO_PATH=(<path to your Arduino IDE Folder containing the arduino executable>)`, save the changes
3) Restart terminal so changes apply, or source bashrc with `. ~/.bashrc`
4) Go to `robotics-prototype/robot/rover`
5) `mkdir build`
6) cd `build`
7) `cmake ..`
If it complains about a missing `WireIMXRT.cpp` file, it is strongly recommended to simply uninstall your arduino ide (`rm the folder`) and reinstall the newest version as well as reinstall teensyduino. Steps to do so are shown below. Anoter quick and dirty unrecommended but tested solution is to copy the contents from [github](https://github.com/PaulStoffregen/Wire/blob/master/WireIMXRT.cpp) and create the file in the Wire folder.
8) `make`

If you want to delete everything that was generated by CMake, including the directories, make sure that the `pwd` is `rover/build`

`cmake --build . --target cleanup`

The only remnant is the empty `build` folder in `robot/rover`

#### Installing arduino IDE + teensyduino from CLI

1. `export ARDUINO_PATH="$HOME/arduino_ide"`
2. `export ARDUINO_IDE_VERSION="1.8.12"`
3. `export TEENSYDUINO_VERSION="152"`
4. `wget https://downloads.arduino.cc/arduino-$ARDUINO_IDE_VERSION-linux64.tar.xz`
5. `mkdir -p $ARDUINO_PATH`
6. `tar xf arduino-$ARDUINO_IDE_VERSION-linux64.tar.xz -C $ARDUINO_PATH`
7. `curl -fsSL https://www.pjrc.com/teensy/td_$TEENSYDUINO_VERSION/TeensyduinoInstall.linux64 -o TeensyduinoInstall.linux64`
8. `chmod +x TeensyduinoInstall.linux64`
9. `./TeensyduinoInstall.linux64 --dir=$ARDUINO_PATH/arduino-$ARDUINO_IDE_VERSION`
10. `cd ~/arduino_ide/arduino-1.8.12` and `./arduino`
11. Once the IDE has opened, verify that you see options for "Teensy" under `Tools -> Board"

closes #204 

# Reviewer Section


robot/rover/Arm/Examples/DcExampleTwoMotorsCMakeLists.txt
```diff
@@ -0,0 +1 @@
+fuck you
```
MartensCedric, line 1: HMMMMMMMMMMMMMMMMMMMMMMMMMMMM

robot/rover/Arm/Examples/RosCommunicationCMakeLists.txt
```diff
@@ -0,0 +1 @@
+fuck you
```
MartensCedric, line 1: please remove

robot/rover/Arm/Examples/ServoExampleOpenLoopCMakeLists.txt
```diff
@@ -0,0 +1 @@
+fuck you
```
MartensCedric, line 1: lmao

robot/rover/CMakeLists.txt
```diff
@@ -0,0 +1,203 @@
+cmake_minimum_required(VERSION 3.5.0)
+set(CMAKE_SYSTEM_PROCESSOR arm)
+
+#execute a shell script to find the arduino directory. It uses the setup.sh file as every arduino version has that in there in the 1st level dir
+execute_process(
+COMMAND find / -name "arduino-linux-setup.sh" -print -quit
+OUTPUT_VARIABLE ARDUINO_SAMPLE_FILE
+OUTPUT_STRIP_TRAILING_WHITESPACE
+)
```
MartensCedric, line 9: pretty smart

MartensCedric: What happens if it doesnt find it? Could you print a message and exit? What happens if they have the ARDUINO IDE installed but not TeensyDuino?

MartensCedric: In retrospect, I think we should keep all the error handling for another PR because there's a lot of that, especially with making sure all libs are there etc.

MartensCedric: A suggestion could be to have a config file that the user just puts their path.

robot/rover/MobilePlatform/dev/path_send_testCMakeLists.txt
```diff
@@ -0,0 +1 @@
+fuck you
```
MartensCedric, line 1: can you remove the fuck yous and maybe put something different or just remove the file

robot/rover/MobilePlatform/dev/pure_pursuit_followerCMakeLists.txt
```diff
@@ -0,0 +1 @@
+fuck you
```
MartensCedric, line 1: Honestly I think some of these dev projects should just be removed @vashmata @PeterGhimself

robot/rover/TestCode/MultipleConstructors/EncoderFuncCMakeLists.txt
```diff
@@ -0,0 +1 @@
+fuck you
```
MartensCedric, line 1: f

robot/rover/TestCode/MultipleConstructorsCMakeLists.txt
```diff
@@ -0,0 +1 @@
+fuck you
```
MartensCedric, line 1: f

MartensCedric:
AH YES, I have 2 arduino installs and it found the non-teensy one lmao

MartensCedric:
> image
> 
> Unfortunately, it didnt build properly for me. I don't know what went wrong

I was just missing a library, I added it to the issue

MartensCedric:
Where is the compile code tho. Like the stuff we need to upload and run on our teensy

MartensCedric:
> Where is the compiled code tho. Like the stuff we need to upload and run on our teensy

Ok I found it, its next to the source in a build file, an ELF

MartensCedric:
I compiled and it worked!

robot/rover/CMakeLists.txt
steamypotato: My previous design had it so that a user entered his own path in the command line, like this : `cmake .. -DARDUINO="...path...."` would this idea be better? its just that its more work for the user I find

steamypotato: I can also make it so that it only finds a Teensy folder, otherwise it lets the user knows, I just assumed everyone already had the Teensy one installed

MartensCedric: hmm if we set our ARDUINO_PATH in ~/.bashrc then the command would be

cmake .. -DARDUINO="$ARDUINO_PATH"

further we could just alias this to alias Build Robotics (BR)

```
alias br='cmake .. -DARDUINO="$ARDUINO_PATH"'
```

MartensCedric: then you just need to type `br` but idk if there's other suggestions @PeterGhimself @vashmata

steamypotato: actually if you set the path in ~/.bashrc , CMake can read env variables right from there, so thats also an option, so it would still just be `cmake ..`

steamypotato: Oh and for the libraries, I didn't add error handling code because there's most likely gonna be a scripted installed of those libraries in an issue, so that would handle any potential errors

steamypotato:
@MartensCedric i noticed you needed to add a different Adafruit library to compile, which is crazy since you already had like 10 Adafruits installed (from your screenshot in #362) , so I assume this problem would happen to anyone else who doesn't have that specific library installed, even though that should be the 'correct' library. Though I could make it so that it searches all installed similar 
libraries and finds the 'correct one' , from a practical standpoint it makes 0 sense to have a bunch of non-used libraries installed, which would likely be a feature of the C++ refactoring issue which would include a scripted install of the necessary libraries, so idk if its worth spending the time to add an auto-find feature, what do you think?

MartensCedric:
It's possible those libraries were from personal projects. I wouldn't do
anything about it for this PR but eventually we can just have a script that
if you don't have the required libraries, it downloads them

robot/rover/CMakeLists.txt
steamypotato: @MartensCedric Okay 2 options for this
- Find the correct Teensyduino folder, even if he has multiple Arduinos installed (like you did), and make sure to let the user know if he doesn't have the proper one installed
- Let the user enter the full path to Teensyduino in bashrc, this is also much easier to code, and more 'safe', as you can't really mess it up from the Cmake side as long as the user did their job right (it would let them know if the path is incorrect)

What's better?

MartensCedric:
Option 2 is easier to code and easy for everyone I think. For now assume
that if the user has the variable defined in bashrc, we will assume that
all the required libraries are installed and teensyduino is installed
correctly. In another PR we will automatically install missing libraries if
that happens. What do you think?

steamypotato:
@MartensCedric yeah thats the 'template' for each CMakeLists.txt file, each sub-project gets its own version using the template

MartensCedric:
I added this wiki page : https://github.com/space-concordia-robotics/robotics-prototype/wiki/Setting-up-Teensyduino


And this one too : https://github.com/space-concordia-robotics/robotics-prototype/wiki/Building-the-C---code

NinjaWeeaboo:
@steamypotato I just tested this and everything compiled for me too!

steamypotato:
> @steamypotato I just tested this and everything compiled for me too!

E P I C C

MartensCedric:
> I slightly updated the PR testing steps.
> 
> So far while testing, step 7 is failing (`cmake ..`), which is strange since I have confirmed that the library seems to be in the right place. I'm using arduino version 1.8.8.
> 
> for_nik_5
> 
> for_nik_6
> 
> I tried to debug, and it seems that the library is being picked up, maybe there's a more thorough check that can be made?
> debug_attempt_0
> 
> @vashmata can you please test this PR and see if it works for you?

Can you go in the wire directory (with the cpp file), then do `pwd` and `ls`. Because this issue arrives when it can't find the cpp file, and if you look online its a path error.

MartensCedric:
FYI it doesn't find the `.cpp` file (as it says in the error), so showing this image 
image

Doesn't mean that it's correctly there. Is the cpp in that folder?

steamypotato:
> FYI it doesn't find the `.cpp` file (as it says in the error), so showing this image
> image
> 
> Doesn't mean that it's correctly there. Is the cpp in that folder?

That's test on 1.8.8 where as you tested on 1.8.9, so it seems 1.8.9 is the minimum needed version

MartensCedric:
Perhaps 1.8.9 is needed, but what I mean is that his grep just checks if
the wired folder exists, he doesn't check if the cpp file is in there,
which would be useful to know

PeterGhimself:
Thanks for pointing that out @MartensCedric, yeah I don't have `WireIMXRT.cpp` anywhere in those folders. I found it online [here](https://github.com/PaulStoffregen/Wire/blob/master/WireIMXRT.cpp), I'm wondering if I can get away with simply copying the file over to the Wire folder manually :thinking:

PeterGhimself:
Well, it's dirty, but copying over the `WireIMXRT.cpp` file manually allowed `cmake ..` step to proceed without errors. Maybe because I setup teensyduino so long ago it didn't include that one particular file.. would still be good to have @vashmata try this since we both installed Teensyduino around the same time. :eyes:

PR #403, opened 2020-06-13
Removed Useless C++ Files

### LINKED ISSUE #377: Remove useless C++ files
We have files like "ideas.h" and "notes.h". There's folders with example code which we will clearly never use. Looking through this document might be useful : https://docs.google.com/document/d/1MZaEAh4fQXDNMqdmfaYagB0u23k_reNE0J9pfIDxf2A/edit#

The assignee must remove as much useless files, projects and example projects as he can in the C++ code.

MartensCedric:
# Assignee Section

## Description
I removed a bunch of useless C++ files such as `ideas.h` or `notes.h` as well as a bunch of test projects. This makes the repo much cleaner.

### Steps for Testing
Ensure that you have the arduino IDE setup from the wiki and can build with cmake.
1. In rover/, do `mkdir build`
2. `cd build`
3. `cmake ..`
4. `make`

Everything still builds because those files were indeed, useless

closes #377

# Reviewer Section


robot/rover/README.md
```diff
@@ -6,9 +6,6 @@ The code running on the rover
 ### test.py
```
PeterGhimself, line 6: change the name to something like `comms_test.py`

PeterGhimself:
CMakeLists.txt also needs to be updated, since as it is it skips over almost all if not all of the projects you deleted here

PeterGhimself:
Ok so by renaming it to `comms_test.py` it got picked up by `pytest` as a unit test, so maybe make it `comms_testy.py`. You can test that locally by changing the name, then running `pytest` from root of repo, rather than having travis test it for you ;)

PR #415, opened 2020-06-15
Update formatters and their configs

### LINKED ISSUE #401: Have properly configured formatter
We are currently using `yapf` and `pylint`

We have these configs 

https://github.com/space-concordia-robotics/robotics-prototype/blob/master/.pylintrc

https://github.com/space-concordia-robotics/robotics-prototype/blob/master/.style.yapf

The assignee should make sure that everything is configured correctly for us to easily be able to run it. The assignee should also document and test the setup.

AFknUsername:
# Assignee Section

## Description
I removed the forced version from requirements-dev.txt for pylint and yapf as it was causing issues (the packages were just too old to work)
I updated some settings in the configuration files to work better with our code base.
Both programs are callable using `pylint` or `yapf`
Keep in mind that yapf outputs a formatted version into sdout, so it doesnt actually change the file itself.

### Steps for Testing
1. Run pylint on some file (I like commit-message-hook.py) and check the messages.
2. Make some changes to the file to try and fix the messages (watch the score go up)
3. Run yapf on some file that is poorly formatted and check the output i guess
4. you could also run pylint on a file, then yapf it and then run pylint again to see what it does.

closes #401 

# Reviewer Section

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -16,10 +16,10 @@ persistent=yes
 
 # List of plugins (as comma separated values of python modules names) to load,
 # usually to register additional checkers.
-load-plugins=
+#load-plugins=
 
 # Use multiple processes to speed up Pylint.
-jobs=1
+jobs=0
 
 # Allow loading of arbitrary C extensions. Extensions are imported into the
 # active Python interpreter and may run arbitrary code.
@@ -28,7 +28,7 @@ unsafe-load-any-extension=no
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code
-extension-pkg-whitelist=
+#extension-pkg-whitelist=
 
 # Allow optimization of some AST trees. This will activate a peephole AST
 # optimizer, which will apply various small optimizations. For instance, it can
@@ -43,7 +43,7 @@ optimize-ast=no
 
 # Only show warnings with the listed confidence levels. Leave empty to show
 # all. Valid levels: HIGH, INFERENCE, INFERENCE_FAILURE, UNDEFINED
-confidence=
+#confidence=
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
@@ -130,7 +130,7 @@ output-format=text
 files-output=no
 
 # Tells whether to display a full report or only the messages
-reports=yes
+reports=no
 
 # Python expression which should return a note less than 10 (10 is the highest
 # note). You have access to the variables errors warning, statement which
@@ -147,7 +147,7 @@ evaluation=10.0 - ((float(5 * error + warning + refactor + convention) / stateme
 [FORMAT]
 
 # Maximum number of characters on a single line.
-max-line-length=100
+max-line-length=200
 
 # Regexp for a line that is allowed to be longer than the limit.
 ignore-long-lines=^\s*(# )?<?https?://\S+>?$|^.*https?://\S+>?$
--- a/.style.yapf
+++ b/.style.yapf
@@ -10,7 +10,7 @@ align_closing_bracket_with_visual_indent=False
 #        'this is the second element of a tuple'):
 #            value,
 #   }
-allow_multiline_dictionary_keys=False
+allow_multiline_dictionary_keys=True
 
 # Allow lambdas to be formatted on more than one line.
 allow_multiline_lambdas=False
@@ -56,7 +56,7 @@ blank_line_before_nested_class_or_def=True
 coalesce_brackets=True
 
 # The column limit.
-column_limit=100
+column_limit=200
 
 # The style for continuation alignment. Possible values are:
 #
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,4 @@
 # Development specific requirements
 pytest==3.6.3
-pylint==2.0.1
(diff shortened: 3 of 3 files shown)


requirements-dev.txt
```diff
@@ -1,4 +1,4 @@
 # Development specific requirements
 pytest==3.6.3
-pylint==2.0.1
-yapf==0.22.0
+pylint
+yapf
```
MartensCedric, line 4: I think we should have fixed versions, its always a pain in the ass if we end up having different versions of software and we end up with different bugs or something because we have a different setup. Given that this goes in the person's `venv` then I think it should def be version specific

MartensCedric: You say in the PR you removed the versions because they were too old, but I would suggest you just pin them back to something newer

AFknUsername: ok, but it will go outdated at some point and break

MartensCedric:
> I removed the forced version from requirements-dev.txt for pylint and yapf as it was causing issues (the packages were just too old to work)

Fake news

MartensCedric:
> ### Steps for Testing
> 1. run `pip install -r requirements-dev.txt`
> 2. Go into the mcu node with `mcunode`
> 3. Run `yapf PdsNode.py >> newfile.py`
> 4. Run `pylint PdsNode.py` and check the score
> 5. Run `pylint newfile.py` and check improved score
> 6. (Optional) Mess around in PdsNode.py to try to resolve issues and check with pylint again to see the score go up.
> 7. Use `git reset --hard HEAD^` to get rid of changes

NICE STEPS

- Peter G

MartensCedric:
> > ### Steps for Testing
> > 
> > 1. run `pip install -r requirements-dev.txt`
> > 2. Go into the mcu node with `mcunode`
> > 3. Run `yapf PdsNode.py >> newfile.py`
> > 4. Run `pylint PdsNode.py` and check the score
> > 5. Run `pylint newfile.py` and check improved score
> > 6. (Optional) Mess around in PdsNode.py to try to resolve issues and check with pylint again to see the score go up.
> > 7. Use `git reset --hard HEAD^` to get rid of changes
> 
> NICE STEPS
> 
-  Peter G

AFknUsername:
> > I removed the forced version from requirements-dev.txt for pylint and yapf as it was causing issues (the packages were just too old to work)
> 
> Fake news

what where?  👀

.pylintrc
```diff
@@ -147,7 +147,7 @@ evaluation=10.0 - ((float(5 * error + warning + refactor + convention) / stateme
 [FORMAT]
 
 # Maximum number of characters on a single line.
-max-line-length=100
+max-line-length=200
```
MartensCedric, line 150: PEP8 requires a max line length of 79 characters tho

AFknUsername:
> > > > I removed the forced version from requirements-dev.txt for pylint and yapf as it was causing issues (the packages were just too old to work)
> > > 
> > > 
> > > Fake news
> > 
> > 
> > what where?
> 
> You didnt remove the force versions, you updated them 😎

Where do you see that? 👀

.style.yapf
```diff
@@ -56,7 +56,7 @@ blank_line_before_nested_class_or_def=True
 coalesce_brackets=True
 
 # The column limit.
-column_limit=100
+column_limit=200
```
MartensCedric, line 59: what is this? I couldnt find online what the difference this was with the max-line-length

MartensCedric:
Can we look into using Black (https://black.readthedocs.io/en/stable/) because if we have an automated formatter, it should be a lifetime ticket to save you for the rest of your life from any formatting change requests. So yapf clearly doesnt have all the features, what about black? The guy that commented today on the link I sent in the previous comment said to perhaps use both

.style.yapf
AFknUsername: It seems like yapf uses column_limit instead of max-line-length

.pylintrc
AFknUsername: .

requirements-dev.txt
AFknUsername: done

.style.yapf
AFknUsername: .

AFknUsername:
> Can we look into using Black (https://black.readthedocs.io/en/stable/) because if we have an automated formatter, it should be a lifetime ticket to save you for the rest of your life from any formatting change requests. So yapf clearly doesnt have all the features, what about black? The guy that commented today on the link I sent in the previous comment said to perhaps use both

Looks like black forces double quotes by default... Maybe we can just not focus on the quotation marks?
👀 @PeterGhimself

AFknUsername:
It seems like its PEP8 or nothing, so I figured that we don't even need configuration files since both packages use PEP8 by default. 
Black seems nicer to use than yapf. 
It is a bit opinionated, but very sensible and uses PEPs to back its choices (Like [PEP257](https://www.python.org/dev/peps/pep-0257/)).

At this point it goes into the opinion territory, so I think we should take what have and use it.

AFknUsername:
@PeterGhimself 
We can def keep the config files in the repo. But maybe you'd care to know that you can always dump the config into stdout or into a file.

Although it should be noted:
yapf seems to be a good balance between black (opinionated formatting, strict guidelines, non configurable) and autopep8 (loose, cares only about pep8 guidelines and not formatting). i.e. **yapf is strict on guidelines, is configurable, and formats the code**

PR #419, opened 2020-06-26
Cmake upload teensy 382

### LINKED ISSUE #382: Add upload to teensy option in CMAKE or make a script for it
The CMake script builds the C++. It should have an option to upload it to a teensy, or we should have a script somewhere that does just that. I think @PeterGhimself had some work from SOEN422, maybe he can comment it here.

steamypotato:
# Assignee Section

## Description
Using CMake, upload the compiled C++ to a Teensy board, using the Teensy Loader CLI program from https://github.com/PaulStoffregen/teensy_loader_cli . It uploads one project at a time, although since the Rover uses 3 Teensys, this can easily be expanded to upload everything at once. Requires a Teensy board to test.

### Steps for Testing
1. Ensure that `rover/config/.bash_aliases` is being sourced, preferably from `~/.bashrc`
2. Go to `robot/rover`, and run `./install_teensy_loader_cli.sh`
3. In `rover`, run `mkdir build && cd build`
4. `cmake ..`
5. `make`
6. Connect your Teensy3.6 board to your computer via usb

#### Now we test every project individually, you can also use a serial monitor to verify data transmission

7. `upload science`
8. `upload Arm`
9. `upload MobilePlatform`

#### You can remove everything that was generated by cmake with
10. `cmake_clean`

closes #382

# Reviewer Section

### DIFF
--- a/robot/rover/config/.bash_aliases
+++ b/robot/rover/config/.bash_aliases
@@ -16,9 +16,12 @@ alias arm="cd $ROVER/ArmDriverUnit"
 alias wheels="cd $ROVER/MobilePlatform"
 alias rostings="cd $ROSPACKAGES"
 alias mcunode="cd $ROSPACKAGES/src/mcu_control/scripts"
+alias upload="cmake --build $ROVER/build --target "
+alias cmake_clean="cmake --build $ROVER/build --target cleanup && rm -rf $ROVER/build"
 
 # edit thyself
 alias editalius="nano $BASH_A"
 alias alius=". $BASH_A"
 alias editnani="nano $NANORC"
 alias nani=". $NANORC"
+
--- a/robot/rover/CMakeLists.txt
+++ b/robot/rover/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.5.0)
+
 set(CMAKE_SYSTEM_PROCESSOR arm)
 include(ExternalProject)
 
@@ -19,6 +20,8 @@ set(ARDUINO_EXTERNAL_LIB /home/$ENV{USER}/Arduino/libraries)
 set(ARM_TOOLS ${ARDUINO_PATH}/hardware/tools)
 set(ARM_LIB ${ARM_TOOLS}/arm/arm-none-eabi/lib)
 set(ARM_GCC_OBJCOPY ${ARM_TOOLS}/arm/bin/arm-none-eabi-objcopy)
+set(TEENSY_LOADER_CLI ${ARM_TOOLS}/teensy_loader_cli)
+
 
 #given a name, will check if the library exists in the proper directory, otherwise run install_arduino_libraries.sh which will install it
 function(checkLibraries lib_name lib_version lib_url)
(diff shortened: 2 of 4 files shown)


MartensCedric:
hot

robot/rover/config/.bash_aliases
```diff
@@ -16,9 +16,12 @@ alias arm="cd $ROVER/ArmDriverUnit"
 alias wheels="cd $ROVER/MobilePlatform"
 alias rostings="cd $ROSPACKAGES"
 alias mcunode="cd $ROSPACKAGES/src/mcu_control/scripts"
+alias upload="cmake --build $ROVER/build --target "
```
MartensCedric, line 19: what exactly makes the upload trigger from this command versus doing just `cmake ..`? Is it the --target option? And why is that?

steamypotato: Targets are just a fancy name for binaries, libraries, or encapsulations for a set of custom bash scripts or configurations. When you do `cmake ..` the targets aren't actually built, they just configured to be built by either `make` (which by default does NOT build custom targets (although there are tricks around this), only binaries and libraries are built), but to upload to a teensy board, there's no default way of doing it like compiling a C++ project or something, you need external scripts hence custom commands hence the use of custom targets. For this issue, it's also much more convenient to build targets from terminal whenever you want, hence this command, where `--target NAME` builds the custom `NAME` target that was configured by `cmake..` , and `--build` just specifies that something is going to be built that was configured in the `$ROVER/build` directory.

steamypotato:
> So I've tested this again. There's another issue completely unrelated to your PR which gave me some false information (See issue #420  ).
> 
> Though more proper digging, me and @PeterGhimself noticed something odd : The upload speed seems too fast. When you upload from the arduino IDE. Your Teensy will stop blinking for like 1-2 seconds. We don't see this with "upload X". We noticed that the code uploaded is not the correct one (it just doesnt upload).
> 
> You can open the Arduino IDE, set Teensy 3.6 and the port then do CTRL+SHIFT+M. When the Teensy is plugged if you see stuff like ASTRO ARM this means the arm code is running. If you see stuff like heading/gps info its the MobilePlatform code.
> 
> I ran `./teensy_loader_cli --mcu=mk66fx1m0 -s -v ../../../Programming/robotics-prototype/robot/rover/Arm/build/bin/Arm.hex` and the upload fails. I think what happens is that the teensy loader script fails.

I know its been a while, but can you be a bit more specific about how you got it to fail? When I open the serial monitor from the Arduino IDE everything works fine (including the commands), also how did you get the upload to fail? I typed in the exact command you did and it worked for me

MartensCedric:
ok I will try to do that again, ping me if I forget in the next day or so

MartensCedric (approved):
Cedric, [26.09.20 21:49]
Hey Nik so I spent a good while investigating the teensy issue. In conclusion, everything's "fine".  I did have issues again. I investigated. I think what was happening previously is that after a load sometimes the teensy would shutdown and I wasnt rebooting it properly so it wouldnt respond (wouldnt upload)

Cedric, [26.09.20 21:50]
I'm not 100% sure I understand what was happening but I tested each load and it works just fine

robot/rover/config/.bash_aliases
MartensCedric, line 19: @AFknUsername @RealWilliamWells We might wanna add robot/**rover**/config/.bash_aliases to the ~/.bashrc, rn its only the one in base being sourced. So this alias didnt work for me. Im not gonna request that change in this PR since its outta scope

PeterGhimself (approved):
[very epic](https://www.youtube.com/watch?v=594URuJkmm4) :sunglasses:

PR #432, opened 2020-08-02
Warning 342

### LINKED ISSUE #342: Emphasize the warning when voltage or temperature become in a warned range
Currently, in the GUI, when the battery voltage or the PDS temperatures are in a dangerous range, the operator gets a model popup. However, when the values are close to the danger value, it is in a warning threshold. Other than changing the text to yellow, not much is done. This issue would require the assignee to make it more clear to the operator that it's in the warning threshold. This could be done by changing the background of some component in the navbar. Alternative solutions are also welcome.

hamzashaikh8:
# Assignee Section

## Description
Add emphasis message when battery voltage or temperature is nearing danger zone (yellow) and add warning sound when battery voltage is in danger zone.

- Added js library notify.js
- Added warning pop op message for battery voltage and temperature
- Added warning sound file warning.wav
- Added warningSound function with the responsibility of triggering the warning sound.
- Replaced call to errorSound function with warningSound when battery voltage is in danger zone

### Steps for Testing

- Open Gui
- Call rostopic command for voltage when nearing danger zone (12.5 < voltage< 13.5):rostopic pub -r 0.2 /battery_voltage mcu_control/Voltage 13
The warning pop up message for battery voltage should appear on the top right of the GUI below the battery temperature.
- Call rostopic command for battery temperatures when nearing danger zone (0<= temperature <10 || 75< temperature <=85) : rostopic pub -r 0.2 mcu_control/ThermistorTemps "therm1: 34.0
therm2: 40.0
therm3: 2.0"
The warning pop up for battery temperature should appear on the top right below the leftest battery temperature
- Call rostopic for voltage when in danger zone (voltage<12.5 || voltage> 16.8) : rostopic pub -r 0.2 /battery_voltage std_msgs/Float32 17
Warning Sound should be heard only once alongside an exit pop indicating to either charge or discharge the battery voltage.

closes #342

# Reviewer Section


AFknUsername:
I just tested it an it seems to work, however, the warning message keeps appearing and disappearing.. Is that intentional?
Also when I set a temp to something low (i.e 2), the message says "too hot". Maybe set the warning to something more general.

Unrelated to your work, it seems like lines 253-290 are pretty messy and over-engineered. Maybe we should open an issue to refactor that? Thoughts?

hamzashaikh8:
> I just tested it an it seems to work, however, the warning message keeps appearing and disappearing.. Is that intentional?
> Also when I set a temp to something low (i.e 2), the message says "too hot". Maybe set the warning to something more general.
> 
> Unrelated to your work, it seems like lines 253-290 are pretty messy and over-engineered. Maybe we should open an issue to refactor that? Thoughts?

The message message appearing and disappearing is intentional. The indicated rostopic commands are meant to mimic real-time values, so the voltage and temperature values are being refreshed 5 times per second ("-r 0.2" in the rostopic commands). By removing "-r 0.2" in the rostopic commands (ex: "rostopic pub battery_voltage std_msgs/Float32 13")  the values will be updated once, so the warning message will not keep appearing and disappearing.

RealWilliamWells:
Just tested this and everything works great! Good job! Later on I'll take a look at your code as well.

robot/basestation/static/js/app.js
```diff
@@ -260,3 +261,10 @@ plays an error sound once -> warning will not play if the document has not been
 function errorSound (){
   new Audio(ERROR_SOUND_PATH).play()
 }
+
+/*
+plays an error sound once -> warning will not play if the document has not been interacted with yet
```
MartensCedric, line 266: comment is confusing, it says its error sound but its called warning?

robot/basestation/static/js/roslibjs/ros-helpers.js
```diff
@@ -243,6 +243,7 @@ function initRosWeb () {
   })
   battery_temps_listener.subscribe(function (message) {
     // sets temperatures to two decimal points
+    console.log(message)
```
MartensCedric, line 246: please remove

robot/basestation/static/js/roslibjs/ros-helpers.js
```diff
@@ -227,7 +227,7 @@ function initRosWeb () {
     } else {
       if (voltage < MIN_VOLTAGE + VOLTAGE_WARNING) {
         textColor('#battery-voltage', 'orange')
-
+        $('#battery-temperature').notify('Warning! Voltage is approaching danger value!', 'warn',{position:"bottom"})
```
MartensCedric, line 230: Why does `#battery-temperature` get the **Voltage** warning notification?

hamzashaikh8: #battery-temperature is the Celsius label to the right of the battery logo and below the battery voltage #battery-voltage, not to get confused with the 3 battery temperature logos on the top-right of the page. #battery-temperature gets the voltage because if #battery-voltage receives it then it hides the battery temperature #battery-temperature and other elements

robot/basestation/static/js/roslibjs/ros-helpers.js
```diff
@@ -212,7 +212,7 @@ function initRosWeb () {
 
       if ($('#battery-voltage').attr('acceptable') === '1' && ALERT_ENABLE) {
         $('#battery-voltage').attr('acceptable', '0')
-        errorSound()
+        warningSound()
```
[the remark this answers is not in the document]
hamzashaikh8: Ok, I added error.wav (same sound as warning.wav) and removed error.mp3 and warning.wav. I also redirected ERROR_SOUND_PATH to error.wav, so errorSound now plays the new error sound. I removed warningSound and replaced all calls to warningSound back to errorSound. In essence, errorSound now plays the new sound and I removed warningSound function.

robot/basestation/static/js/app.js
hamzashaikh8: Removed warningSound and redirected errorSound to new sound

robot/basestation/static/js/app.js
```diff
@@ -1,5 +1,6 @@
 PING_THROTTLE_TIME = 1000
-ERROR_SOUND_PATH = '../../static/sound/error.mp3'
+ERROR_SOUND_PATH = '../../static/sound/error.wav'
```
AFknUsername, line 2: I feel like it would make more sense to call this the warning sound.
Is this sound used for actual errors in some places?

robot/basestation/static/js/roslibjs/ros-helpers.js
```diff
@@ -276,12 +276,13 @@ function initRosWeb () {
...
           $obj.css({'color': 'orange'})
-
+          $('#battery-temp-1').notify('Battery temperature hot!', 'warn',{position:"bottom"})
         } else {
           $obj.css({'color': 'white'})
         }
 
-        if ($obj.attr('acceptable') === '0') $obj.attr('acceptable', '1')
+        if ($obj.attr('acceptable') === '0')
+         $obj.attr('acceptable', '1')
```
AFknUsername, line 285: Indentation :fearful:

robot/basestation/static/js/app.js
hamzashaikh8: Yes, this is the sound triggered when the voltage or temperature is in danger zone (red text).

MartensCedric: @AFknUsername There's two levels of notification for the voltage. 

We have a zone that if the voltage is under that value, we want to immediately cease operations because it could damage the rover. Hence, when this value is reached there's an error sounds and a pop up model dialogue. The operator is expected to hit the killswitch of the rover. This is bad though, because we never actually want to enter this danger zone.

This issue was created to notify the operator during a warning zone (not in the error zone but close). The warning sends a light notification (not a popup, but a top-right chrome-styled notification). It is lightweight so it doesnt distract the operator in a critical moment, but still visible. AFAIK it does not have a sound. So the sound is for when we are in the danger zone. @AFknUsername

robot/basestation/static/js/roslibjs/ros-helpers.js
MartensCedric: f

robot/basestation/static/js/roslibjs/ros-helpers.js
MartensCedric: This is a bit confusing, perhaps you should create a parent div with a more appropriate name. Glancing in the code, it looks like an error. I wouldn't be surprised if someone would try to change it to #battery-voltage thinking it was an error

robot/basestation/static/js/roslibjs/ros-helpers.js
MartensCedric: 1 space indentation, new meta lmao

robot/basestation/static/js/app.js
AFknUsername: I see that makes much more sense, thanks

MartensCedric:
## To clear the main confusion for reviewers

From one of the threads, this explains the difference between warning and error zone
> @AFknUsername There's two levels of notification for the voltage.
> 
> We have a zone that if the voltage is under that value, we want to immediately cease operations because it could damage the rover. Hence, when this value is reached there's an error sounds and a pop up model dialogue. The operator is expected to hit the killswitch of the rover. This is bad though, because we never actually want to enter this danger zone.
> 
> This issue was created to notify the operator during a warning zone (not in the error zone but close). The warning sends a light notification (not a popup, but a top-right chrome-styled notification). It is lightweight so it doesnt distract the operator in a critical moment, but still visible. AFAIK it does not have a sound. So the sound is for when we are in the danger zone.

robot/basestation/static/js/roslibjs/ros-helpers.js
hamzashaikh8: Ok I added a new inner div 'battery-voltage-warning' and passed the warning notification through there. The location of the pop up is still the same

robot/basestation/static/js/roslibjs/ros-helpers.js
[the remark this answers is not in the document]
hamzashaikh8: @MartensCedric

robot/basestation/static/js/roslibjs/ros-helpers.js
hamzashaikh8: @MartensCedric

MartensCedric (changes requested):
Hey Hamza, everything looks fine! Although you'll need to make a change to make sure it works with the rover. On the rover, the battery voltage is published A LOT. (Probably around 20 times per second). If I put a rate of 10 Hz, I get this problem where it spams it so fast that we can't even read the notification. I suggest you only pop the message once when it goes below the threshold and dont make it appear everytime it receives a new message under the threshold. But then if the battery voltage goes back up and back down, it should pop back again. I think this should be enough. Eventually we might wanna try to handle noise but this could go in a different issue.

MartensCedric:
ezgif-4-f2489cb439c9

AFknUsername:
I agree with @MartensCedric. Maybe implement a simple check to ignore subsequent notifications if the error/warning state is already active? We don't need it flashing constantly

MartensCedric:
The issue though is that let's say the threshold is 13.5 and the battery
voltage is at 13.5 it will never output 13.5, it will spam values +-0.05V
so you'll get a spam of 13.45 and 13.55 so there's still going to be some
spam lol. Adrien made an algorithm which is why his code is complicated so
that it never spams the danger zone. But we can do that in another issue

AFknUsername:
Oh lmao, that's mildly problematic

robot/basestation/static/js/roslibjs/ros-helpers.js
```diff
@@ -227,8 +239,13 @@ function initRosWeb () {
     } else {
       if (voltage < MIN_VOLTAGE + VOLTAGE_WARNING) {
         textColor('#battery-voltage', 'orange')
+        crntalert = Date.now();
+        console.log("current alert is: " + crntalert);
+        console.log("previous alert is: " + prevalert);
+        console.log(checkAlert() == true);
```
VimFreeman, line 245: Please remove before merging

MartensCedric (changes requested):
Everything works!! Except there's a minor issue :  
image

It should be "Is too cold!!"

robot/basestation/static/js/roslibjs/ros-helpers.js
```diff
@@ -227,7 +239,12 @@ function initRosWeb () {
     } else {
       if (voltage < MIN_VOLTAGE + VOLTAGE_WARNING) {
         textColor('#battery-voltage', 'orange')
-
+        crntalert = Date.now();
+        console.log("current alert is: " + crntalert);
+        console.log("previous alert is: " + prevalert);
```
MartensCedric, line 244: please remove this

MartensCedric (changes requested):

MartensCedric (dismissed):

VimFreeman:
It says temperature "cold" even when it is too hot

MartensCedric (changes requested):
1. When it goes below 13.5 and goes back up it stays red (should become yellow). 
2.  Temperature warning flickers too fast (Do the same fix as voltage)
3. Temperature "Hot" should be something like 55 I guess not 75

RealWilliamWells:
Closing because web GUI isn't used anymore

PR #435, opened 2020-08-12
[#284] Implement local mock parameter publisher for battery_voltage, battery_temps and wheel_motor_currents

### LINKED ISSUE #284: Local PDS Simulator
This will probably be accomplished by creating a bunch of rospy python scripts.
We need to have some "fake" publishers to mock the real data that would be published if the rover actually on.

launch parameters : mode ...

- [x] Write initial publisher rospy script(s)
- - [ ] Temperature
- - - [ ] Stable
- - - [ ] Unstable
- - - - [ ] start stable and go past max
- - - - [ ] start stable and go past min
- - [x] Voltage

ginywiny:
## Description
The user can set parameters for battery_voltage, battery_temps, and wheel_motor_currents. There are 1 voltage parameter, 3 temperature parameters, and 6 current parameters. Each parameter can be set to rise, remain stable, or fall, and each value published for the parameter suffers noise. The parameters reach their assigned states incrementally based on a user defined rate. 

### Steps for Testing
1. Run `rosgui` and `startgui` in two terminal windows simultaneously.
2. Open the GUI on http://localhost:5000.
3. Change directories to `robotics-prototype/robot/util/rosRoverStart` in another terminal window.
4. Run local mock values using `roslaunch local_rover.launch`.
5. Observe the published values either through the same terminal window as the one launching `roslaunch local_rover.launch` or directly through the GUI. 
6. Observe list of parameters to write to using `rosparam list` . Note: The parameters are called PDS_mock_voltage, PDS_mock_temp# (# is 1 to 3), and PDS_mock_wheel#_current (# is 1 to 6).
7. Set a parameter to rise, constant or fall using `rosparam set /<parameter> <state>`. Note: The parameter states are rise, constant and fall. 
8. Run `rostopic echo <parameter>` in another terminal window to see a specific parameter that is published. 
9. Observe that battery_voltage can be set to 20, 15 and 10, battery_temps to 100, 25, -20, and wheel_motor_currents to 1, 0.3 and 0. 
- **Note**: outputs for all parameters have a difference of +- their noise value. **However**, if the lowest value the variable can reach is 0, therefore the value cannot descend below 0. 
9. Can set different float values for the voltage, temperature and current increment/decrement rates and noise intensities.
10. Set a condition or float value in the `roslaunch local_rover.launch` file for each parameter in order for the simulator to produce the written conditions on startup. 

### **Important Notes**
**1.** If the user removes a parameter from the launch file, the simulation sets the value for that parameter to 0.
- **Example:** `"PDS_mock_voltage"` is never set initialized in launch file, then the GUI will display 0. 

**CRUCIAL NOTE WHEN RUNNING WITH MISSING PARAMETER:** If the user attempts to set the uninitialized parameter from CLI using: `rosparam set /"PDS_mock_voltage" <state>`, the CLI will not display the parameter name from the list of rosparams. 
**For example**, `"PDS_mock_voltage"` is NOT seen as an option from the list of rosparams after. 
**However**, if the user already knows the parameter name by heart regardless, and runs the set command with `"PDS_mock_voltage"`, the next time the launch file is started, then `"PDS_mock_voltage"` will appear as an existing parameter from the rosparam list, and the parameter value on the GUI will be set to the state the user set. This is because the user MANUALLY set the parameter to exist, even though it should not. 

- **Example:** `rosparam set /"PDS_mock_voltage" rise` will set `"PDS_mock_voltage"` to `rise` instead of 0 when the user stops and relaunches the launch file. 
- **Solution:** To reset the parameter to non-existent, the user must stop and restart the gui with `rosgui` and `startgui`

Closes #284


robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
```diff
@@ -0,0 +1,283 @@
...
+from mcu_control.msg import ThermistorTemps, Voltage, Currents
+from decimal import *
+
+
+#Log all data during publishing
+#Add to publish_mock_data to log published data in CLI
+#Remove from publish_mock_data if not needed
+#Used for testing purposes
+def rospyPublishedValueLogger(prevParameterStates, prevParameterValues):
+    #Log new values when posted
```
MartensCedric, line 18: FYI this is not how you do pythondoc comments 

https://www.python.org/dev/peps/pep-0257/

MartensCedric: there's other instances of comments not using proper docstrings, I wont comment on them but this implies you should change them all

ginywiny: I applied a change, not 100% sure if it is right though, but let me know once you see the update.

MartensCedric: Yeah, except that the comments are actually UNDER the function in python lol. You'll eventually get used to it

MartensCedric:
Btw I think you're github email isnt setup correctly because look at the commits on Github. Usually its because `git config --global user.email` isn't set to the correct email (Should be your GH email)

robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
ginywiny: Whoops yeah, i wasn't too sure when i read the docs. I'll change this. Bruh moment

ginywiny:
> Btw I think you're github email isnt setup correctly because look at the commits on Github. Usually its because `git config --global user.email` isn't set to the correct email (Should be your GH email). And user.name should be your GH name

I accidentally set my SpaceConcordia email as opposed to my GH account email. I have changed this since you added your comment. Let me know in case you still see an issue with it.

robot/util/rosRoverStart/local_rover.launch
```diff
@@ -2,4 +2,15 @@
...
+    <param name="PDS_mock_voltage" value="stable"/>
+    <param name="PDS_mock_temp1" value="stable"/>
+    <param name="PDS_mock_temp2" value="stable"/>
+    <param name="PDS_mock_temp3" value="stable"/>
+    <param name="PDS_mock_wheel_current1" value="stable"/>
+    <param name="PDS_mock_wheel_current2" value="stable"/>
+    <param name="PDS_mock_wheel_current3" value="stable"/>
+    <param name="PDS_mock_wheel_current4" value="stable"/>
+    <param name="PDS_mock_wheel_current5" value="stable"/>
+    <param name="PDS_mock_wheel_current6" value="stable"/>
```
MartensCedric, line 15: A few things could be added here.

- The starting value (regardless if you are stable, rising or falling, you should start with the same value. Battery should be 15V, temps 25 and currents I actually have no clue ). The starting value can be 1 param for voltage, 1 for all temps and 1 for all currents.

- Noise value should be a ROS param. I don't think you need a noise value per param. Maybe just make 1 noise value param that applies for everything. `0.05` is a good default imo so that's good

- Rise/Fall rate should also be a param. The rate can be 1 param for voltage, 1 for all temps and 1 for all currents.

robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
```diff
@@ -0,0 +1,199 @@
...
+    currentTemps = []
+    parameterTemps = ["PDS_mock_temp1", "PDS_mock_temp2", "PDS_mock_temp3"]
+    currentTemps = get_states_values(parameterTemps, temps)
+    rospy.logwarn("currentTemps: {}".format(currentTemps))
+
+    #Initialize 6 starting wheel currents
+    currentWheelCurrents = []
+    parameterCurrents = ["PDS_mock_wheel1_current", "PDS_mock_wheel2_current", "PDS_mock_wheel3_current", "PDS_mock_wheel4_current", "PDS_mock_wheel5_current", "PDS_mock_wheel6_current"]
+    currentWheelCurrents = get_states_values(parameterCurrents, currents)
+    rospy.logwarn("currentWheelCurrents: {}".format(currentWheelCurrents))
```
MartensCedric, line 120:
```suggestion
    rospy.loginfo("currentRate: {}".format(currentRate))

    #Initialize 1 starting noise
    currentNoise = get_states_values("PDS_mock_global_noise", noiseValues)
    rospy.loginfo("currentNoise: {}".format(currentNoise))

    #Initialize 1 starting voltage
    currentVoltage = get_states_values("PDS_mock_global_noise", voltages)
    rospy.loginfo("currentVoltage: {}".format(currentVoltage))

    #Initialize 3 starting temperatures
    currentTemps = []
    parameterTemps = ["PDS_mock_temp1", "PDS_mock_temp2", "PDS_mock_temp3"]
    currentTemps = get_states_values(parameterTemps, temps)
    rospy.loginfo("currentTemps: {}".format(currentTemps))

    #Initialize 6 starting wheel currents
    currentWheelCurrents = []
    parameterCurrents = ["PDS_mock_wheel1_current", "PDS_mock_wheel2_current", "PDS_mock_wheel3_current", "PDS_mock_wheel4_current", "PDS_mock_wheel5_current", "PDS_mock_wheel6_current"]
    currentWheelCurrents = get_states_values(parameterCurrents, currents)
    rospy.loginfo("currentWheelCurrents: {}".format(currentWheelCurrents))
```

image

Me after seeing all these warnings 
image

MartensCedric: tagging @AFknUsername because he will :eye: the meme and lol from his phone. then maybe review the PR

MartensCedric: btw @ginywiny when you accept this don't forget to add the "[#284]" because it doesnt use the hook

robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
```diff
@@ -0,0 +1,199 @@
...
+if __name__ == '__main__':
+    """Calls publisher function and passes parameter settings
+    Change the values for the parameter states (rise, stable and fall) here (i.e. voltages, temps and currents)
+    No other places require to be modified to change the state values"""
+
+    #Set voltages rise, stable and fall to be parametrized
+    voltages = {"rise": 20, "stable": 15, "fall": 10}
+
+    #Set temperatures rise, stable and fall to be parametrized
+    temps = {"rise": 100, "stable": 50, "fall": -20}
```
MartensCedric, line 184:
```suggestion
    temps = {"rise": 100, "stable": 25, "fall": -20}
```

robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
```diff
@@ -0,0 +1,199 @@
...
+        rospy.logwarn("Parameter state None assigned to {}.".format(parameter))
+        parameterState = None
+    return parameterState
+
+def publish_mock_data(voltages, temps, currents, noiseValues, riseFallRates):
+    """Get and set parameters to be published"""
+
+    #Initialize node
+    node_name = "mock_pds_node"
+    rospy.init_node(node_name, anonymous=True)
```
MartensCedric, line 85:
```suggestion
    rospy.init_node(node_name)
```

This should prevent people from running 2 instances of this node

ginywiny:
### 1. If the user removes a parameter from the launch file, the simulation sets the value for that parameter to 0.
- **Example:** `"PDS_mock_voltage"` is never set initialized in launch file, then the GUI will display 0. 

**IMPORTANT NOTE:** If the user attempts to set the uninitialized parameter from CLI using: `rosparam set /"PDS_mock_voltage" <state>`, the CLI will not display `"PDS_mock_voltage"` as an option from the list of rosparams. **However**, if the user already knows the parameter name by heart regardless, and runs the set command with `"PDS_mock_voltage"`, the next time the launch file is started, then `"PDS_mock_voltage"` will appear as an existing parameter from the rosparam list, and the parameter value on the GUI will be set to the state the user set. 
- **Example:** `rosparam set /"PDS_mock_voltage" rise` will set `"PDS_mock_voltage"` to `rise` instead of 0 when the user stops and relaunches the launch file. 
- **Solution:** To reset the parameter to non-existent, the user must stop and restart the gui with `rosgui` and `startgui`

### 2. The user is now capable of setting the `rise/fall rate` and `noise` value to 3 different possible values (rise, fall, stable).

### 3.  The rise/fall rate has decreased. 

### 4. The code has been refactored for better readability.

robot/util/rosRoverStart/local_rover.launch
```diff
@@ -2,4 +2,17 @@
     <node pkg="mux_selector" type="mux_select_server.py" name="mux_select_server" output="screen" args="local"/>
     <node pkg="web_video_server" type="web_video_server" name="web_video_server" output="screen"/>    
     <node pkg="task_handler" type="task_handler_server.py" name="task_handler_server" output="screen" args="local"/>
+    <node pkg="mcu_control" type="MockPdsNode.py" name="mock_pds_node" output="screen"/>
+    <param name="PDS_mock_rise_fall_rate" value="stable"/>
```
MartensCedric, line 6: what is this

MartensCedric: why is it "stable"? Shouldnt it be a float?

MartensCedric (changes requested):

robot/util/rosRoverStart/local_rover.launch
ginywiny: Im reusing the 3 possible states for the rise/fall rate (rise = 0.1, stable = 0.04, and 0.01). Unless you want the user to be able to manually enter a value of their liking.

MartensCedric: when I tested to "rise" temp1 it would go to 100 instantaneously

ginywiny: Fixed this

MartensCedric (changes requested):
Voltage and temperatures work great. I investigated the current issue, and its **not** a GUI bug. 
image
This is the values being sent monkaS

robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
```diff
@@ -0,0 +1,276 @@
...
+    voltages = {"rise": 20, "stable": 15, "fall": 10}
+
+    #Set temperatures rise, stable and fall to be parametrized
+    temps = {"rise": 100, "stable": 25, "fall": -20}
+
+    #Set temperatures rise, stable and fall to be parametrized
+    currents = {"rise": 1, "stable": 0.3, "fall": 0}
+
+    #Set temperatures rise, stable and fall to be parametrized
+    noiseValues = {"rise": 0.1, "stable": 0.05, "fall": 0.01}
```
MartensCedric, line 270: this doesnt make sense as we discussed. The fact that the noise is small isnt falling. The noise value should be a float not a string in the parameters. Also can we have noise for voltage, noise for currents and noise for temps. Because a noise of 0.05 is huge for currents lol but its appropriate for temps and voltage

MartensCedric (changes requested):

robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
ginywiny: Ah yes, i completely forgot about that. So i take noiseVoltage, noiseCurrent, noiseTemps, and all of which are set using a float as opposed to a state (rise, stable, fall).

ginywiny:
> Voltage and temperatures work great. I investigated the current issue, and its **not** a GUI bug.
> image
> This is the values being sent monkaS. You're sending like 1000 values lol

Wow, this actually makes sense. Completely went over my head, I have a fix for this.

MartensCedric:
> > Voltage and temperatures work great. I investigated the current issue, and its **not** a GUI bug.
> > This is the values being sent monkaS. You're sending like 1000 values lol
> 
> Wow, this actually makes sense. Completely went over my head, I have a fix for this.

I already fixed it. I pushed a commit

robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
MartensCedric: very cool

MartensCedric:
before merging, we should rebase to edit your email that you used which didnt register with a git account

image

robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
ginywiny: fixed

MartensCedric (changes requested):
Voltage and temperature noise is using 0.01

MartensCedric (dismissed):

AFknUsername (dismissed):
Testing steps need to be much clearer.
I could not reproduce the steps so I was not able to view the topic echo for any mock topic.

AFknUsername:
With some help from @PeterGhimself I got the temps and current to rise and fall, and the state is accurately reflected with `rosparam get /<parameter>`
However, the voltage does not seem to drop when set to "fall"

MartensCedric:
tim, make sure you kill everything. The simulation isn't required to
dynamically change as the scenarios are gonna be static and launched with a
launch file

MartensCedric:
tldr restart the thing with "fall" in the launch file

AFknUsername:
@MartensCedric ok, then everything seems good

PeterGhimself (changes requested):
I tested all the different types (voltage, wheel_current, temps), it all seemed to work according to the plan, except temperatures. they all stay at around 25 degrees for me no matter what I set for the rosparam.

Btw, I used `rosparam set PDS_<etc> <rise|fall|stable>` while running the local_rover.launch file in the background. 

A few things:
- style formatting is not compliant with PEP8, see my inline comments for more info (`pylint` is your friend)
- can you change the default rate of the published values? just cause on the real rover the values don't fluctuate that frequently. I'd say the changes occur about once per second on average. Other than that it's a nice feature that makes it seem more realistic.
- The default amount of time it takes to ramp up/down to the target value is kinda slow. Can we pass a parameter to set a target time frame? like if I wanted it to rise within the period of 2 seconds rather than whatever the default is (definitely longer than 2 seconds). Maybe this is a bit overkill though.
- Since the minimum of the currents is 0, and since there is a +/- of 0.05, it shows the value as negative when issued `fall` parameter... not sure if this is realistic perhaps we can get @TheRealJeff 's opinion on that.
- Plz add some documentation on this in [here](https://github.com/space-concordia-robotics/robotics-prototype/blob/master/robot/rospackages/README.md). You could probably reuse some stuff from your PR description (which I have edited a bit) to speed up the process.

robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
```diff
@@ -0,0 +1,287 @@
+#! /usr/bin/env python3
+
+import time
+import rospy
+import sys
+import random
+from std_msgs.msg import Float32
+from geometry_msgs.msg import Point
+from mcu_control.msg import ThermistorTemps, Voltage, Currents
+from decimal import *
```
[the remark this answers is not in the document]
ginywiny: Thank you for the comments! 
1. I'll look into PEP8 to reformat!

2. It is possible to change the `rospy.Rate` to something more appropriate if the values are fluctuating too quickly. 

3. How long have you let the temperature to rise/fall? It seems to be fine on my setup; however, the rates of rising and decreasing are initially quite slow (`PDS_mock_voltage_rate` set to 0.04, `PDS_mock_temp_rate` set to 1 and `PDS_mock_current_rate` set to 0.025). Because of this, it takes a while for the values to reach their target. Next, the time it takes to rise to a particular value can be decreased by changing the increment rates (`PDS_mock_voltage_rate`, `PDS_mock_temp_rate`, `PDS_mock_current_rate`) as opposed to defining a specific time to reach the target value. This is done by setting a float value, and originally the values are set to 0.04, 1 and 0.025 respectively. However, these values can be made larger so the time to reach the target value is decreased. 

4. I will take a look into the negative value, I can add a condition to avoid this.

5. Would you like me to create a dedicated wiki page for the simulation?

robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
```diff
@@ -0,0 +1,287 @@
...
+    temps = {"rise": 100, "stable": 25, "fall": -20}
+
+    #Set temperatures rise, stable and fall to be parametrized
+    currents = {"rise": 1, "stable": 0.3, "fall": 0}
+
+    #Publish all new parameter settings set by user
+    try:
+        publish_mock_data(voltages, temps, currents)
+    except rospy.ROSInterruptException:
+        pass
```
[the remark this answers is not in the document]
ginywiny: > @MartensCedric ok, then everything seems good

I will be adding more documentation to better understand and run this.

ginywiny:
@AFknUsername 

I will be adding more documentation to better understand and run this. However, is there anything else that was confusing?

AFknUsername:
@ginywiny 
The only thing that I would change is in the testing steps. 
It says to run "roslaunch rover_local.launch", but fails to mention that it must be run in the 
`robotics-prototype/robot/util/rosRoverStart` directory.
That is all

ginywiny:
1. I will be adding [documentation](https://github.com/space-concordia-robotics/robotics-prototype/blob/master/robot/rospackages/README.md); however I would like to be sure of my changes before I proceed.

2. PEP8 is telling me to rename MockPdsNode to snake_case. Should I still do this? I have noticed all other scripts all being written in the same style as MockPdsNode. @PeterGhimself 

3. I have updated the protocol for running the simulator @AFknUsername

ginywiny:
@PeterGhimself @AFknUsername @RealWilliamWells @MartensCedric Let me know if any additional changes should be made.

robot/rospackages/src/mcu_control/scripts/MockPdsNode.py
```diff
@@ -0,0 +1,287 @@
...
+    try:
+        floatValue = float(paramValue)
+    #If the value entered is not a float, keep the value it was set to before
+    except:
+        rospy.logwarn("Invalid parameter set, value unchanged: {}".format(safetyValue))
+        floatValue = float(safetyValue)
+    return floatValue
+    
+
+def get_param_exist(parameter):
```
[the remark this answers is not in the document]
MartensCedric: I think in the future it could be useful but I think this is fine for now but as we have more nodes we should have oen

MartensCedric (approved):

RealWilliamWells (approved):
I briefly looked at your code then tested the publisher and everything seems to work just fine! Good job gamer 😎😎😎

PR #472, opened 2020-10-17
Modularize mobile platform 451

### LINKED ISSUE #451: Modularization of the C++ in Mobile Platform
First thing that will be done is to remove the external references in the header files included in MobilePlatform.ino.
By the end of this issue, every file in MobilePlatform will compile on its own. This issue needs to be completed before #379 can be completed. Note that this is not a complete overhaul of the C++ code, just a starting point for many issues to come (consult SQA docs if you're curious)

steamypotato:
# Assignee Section

## Description
Refactors the mobile platform project in such a way that the includes are no longer in the middle of the .ino file, which also allows the header files to be separated in their respective .h and .cpp files. Also, it removes the TinyGPS library from the project and just installs it like the other libraries.

Two of the main ideas in the refactoring were to turn the helper class into a static singleton and solve the dependency issue by passing object references into the Commands class.

### Steps for Testing
Use cmake to build the project https://github.com/space-concordia-robotics/robotics-prototype/wiki/Building-the-C---code

closes #451

# Reviewer Section


robot/rover/MobilePlatform/includes/Helpers.h
```diff
@@ -0,0 +1,37 @@
...
+
+    static Helpers& get();
+
+    void print(String msg);
+
+    void println(String msg);
+
+    void printres(float msg, int a);
+  
+};
```
MartensCedric, line 34: in terms of software design there's not actually a purpose for this class. Normally a class has a responsibility. It might just be better to actually have all of these function globals. If you can logically group them into something else like ioutils (ser_flush, prints) and other types of utils it might look cleaner. They can still be in a similar namespace like ioutils::ser_flush()

MartensCedric (changes requested):
I'm assuming the goal of this PR was to get ride of the "mid-file" includes, is this correct? (I read the issue and im not 100% sure). Anyway I'm def not gonna request any changes on code you didnt write, I'm not sure when we can test this but I'll try once Koa is back.

MartensCedric:
Note : If you build with the Arduino IDE, the code does not build correctly anymore. Works just fine with the cmake tho

MartensCedric (changes requested):
This is definitely an improvement, I made extra suggestions and tested on my machine. We will eventually test on the rover to validate that nothing broke

MartensCedric (approved):

VimFreeman:
We will approve this PR and deal with the potential consequences later

VimFreeman (approved):

PR #482, opened 2020-12-21
Ar tag detection 481

MewtR:
# Assignee Section

## Description
This pull request integrates the ar tag detection from [here](https://github.com/space-concordia-robotics/docker-erc) to this repository.

#### Prerequisites
- Make sure you install [ar_track_alvar](http://wiki.ros.org/ar_track_alvar): `sudo apt install ros-melodic-ar-track-alvar`
- Calibrate your camera by following the instructions [here](http://wiki.ros.org/camera_calibration/Tutorials/MonocularCalibration)
- Print out some AR tags. They can be found [here](http://wiki.ros.org/ar_track_alvar?action=AttachFile&do=view&target=markers0to8.png) for example.

### Steps for Testing
1. Run the project and visit `http://localhost:5000`.
2. Select a camera stream and start it.
3. Click on the "AR" button, you should see a red square appear.
4. Hold up the printed AR tags in front of you camera. The red square should turn green.
5. Click on the "AR" button again and the square should disappear.

#### Additional Steps
1. Test with multiple streams simultaneously.

closes #481

The approval from all software team leads is necessary before merging.

# Reviewer Section

Aside from local testing and the General Integration Test it is implied that static analysis should be included in the verification process.

- [ ] Local Test Performed Successfully
- [ ] [General Integration Test](https://docs.google.com/document/d/1ug0CpA1cIzURP8DDFSvCt2CEJJSwJ6Ta6B1LG_hYk6I/edit) Performed Successfully

For Pull Requests that do not include code changes, it is not required to perform the tests above.

PR #492, opened 2021-01-02
Ar tag detection manual 484

### LINKED ISSUE #484: Start AR stream only when needed.
This should help with the bandwidth and maybe performance. When you launch the GUI it shouldnt launch every stream in AR node.

MewtR:
# Assignee Section

## Description
Essentially same as #482 with some minor improvements. If this is approved we can actually skip merging #482 and just merge this one instead because it has all the same commits.

Main points:
- The AR stream is only started when a request is made to start it (by clicking on the AR button)
- Clicking the AR button while AR detection is on will only kill the nodes if no one else is currently using it.

#### Prerequisites (from #482)
- Make sure you install [ar_track_alvar](http://wiki.ros.org/ar_track_alvar): `sudo apt install ros-melodic-ar-track-alvar`
- Calibrate your camera by following the instructions [here](http://wiki.ros.org/camera_calibration/Tutorials/MonocularCalibration)
- Print out some AR tags. They can be found [here](http://wiki.ros.org/ar_track_alvar?action=AttachFile&do=view&target=markers0to8.png) for example.

I won't include the steps for testing the actual AR detection (making the red square turn green). Refer to #482 for how to do that.

1. Start the project as usual
2. Start a video stream as usual and click on the `AR` button to get the red square (it takes a few seconds).
3. At this point you should see `/ar_track_alvar_X` as well as `/ar_tracker_X` in the output of `rosnode list`.  (`X` being `video0Cam` or whatever the topic is called)
4. Click the `AR` button again and wait for it to shut down. Go back to the terminal and make sure the AR nodes are no longer in the output of `rosnode list`
5. Click the `AR` button again to turn it back on.
6. Open the project in a new tab and get an AR stream going there as well.
7. Clicking the `AR` button in the first tab tab shouldn't kill the AR nodes but the red square should still disappear.
8. At this point you can play around with turning them on/off in either tab. As long as one of them remains on the AR should nodes should remain alive.
9. Turn off the AR streams in both tabs and watch the AR nodes die again.

closes #484

# Reviewer Section


MewtR:
Refer to #482 for beginning of the discussion.

domludera:
Tested the steps locally, and it works as specified

PeterGhimself:
Same as @domludera, just tested and it works fine on my end

robot/basestation/static/js/elements/feed/camera.js
```diff
@@ -135,7 +154,6 @@ $(document).ready(() => {
     cameraPower.attr('src', POWER_ON)
 
     cameraControls.attr('camera-name', cameraName)
-    console.log("memes")
```
PeterGhimself, line 138: :eyes:

robot/basestation/static/js/elements/feed/camera.js
```diff
@@ -286,7 +305,8 @@ $(document).ready(() => {
 
     let streamURL = getStreamURL(getCameraFilename(cameraPanel) + TOPIC_SUFFIX)
     let isPoweredOn = cameraPower.attr("power-on") == "true"
```
PeterGhimself, line 307: for future reference, since this is definitely beyond the scope of this PR and should really have it's own issue relating to cleaning up messy design from before, I will just mention that there are a lot of occurrences of this bad practice of relying on GUI button attributes to determine the state of real things on the rover.

I would propose that instead we should use more and more topics (though I know there may be some sort of limit, so keeping that in mind) that would essentially publish states like these from the rover side. So then our buttons would subscribe to those topics and react accordingly.

Issue #498, opened 2021-01-09
Define a message format


MartensCedric:
A definition of the message format should be defined. A google docs should be created (This one can be updated : https://docs.google.com/spreadsheets/u/1/d/1bE3h0ZCqPAUhW6Gn6G0fKEoOPdopGTZnmmWK1VuVurI/edit?usp=drive_web&ouid=111054195812684486479)

The message format should tell us exactly what information is contained, the sheet should make sure that we have the information of the command ID of each command, their description, arguments, argument type etc. So that other members can easily refer to that sheet.

ginywiny:
Created a very basic draft just to make a simple idea. Would need others to give their thoughts.

VimFreeman:
I added some comments
We should meet anyways today at some point

ginywiny:
Created the Rover and Science formats with @VimFreeman and @MartensCedric. PDS and arm commands message format have yet to be done.

PR #501, opened 2021-01-09
Created the groundwork for a comms library for internal serial communication used by all Teensies

### LINKED ISSUE #495: C++ Send/Receive Serial messages with specified baudrate in new library shared by teensies
The C++ internal comms library should be able to send and receive messages with a specified baudrate

MartensCedric:
# Assignee Section

## Description
A library called `comms` is now in robot/rover where it can be used by each teensy. Commands can be read, written to serial as well as called from the inherited commandCenter of each teensy.

### Steps for Testing
Travis running successfully is sufficient. This PR is not meant to add functional code but more a skeleton where other members will be able to build code on.

closes #495 

# Reviewer Section


robot/rover/comms/src/Serial.cpp
```diff
@@ -0,0 +1,37 @@
+#include "Serial.h"
+
+namespace internal_comms {
+
+    void startSerial(uint8_t rxPin, uint8_t txPin, long baudRate)
+    {
+        pinMode(rxPin, INPUT);
+        pinMode(txPin, OUTPUT);
```
VimFreeman, line 8: We don't need this because `Serial` defines the pins already.
On teensies we have Serial which is usb and then Serial# where # is the uart port

VimFreeman (changes requested):

robot/rover/comms/src/CommandCenter.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "include/CommandCenter.h"
+
+namespace internal_comms
+{
+    Command* CommandCenter::processCommand() const
+    {
+        uint8_t commandID = Serial.read();
+        size_t bytesToRead = Serial.read();
+        uint8_t* buffer = (uint8_t*) malloc(sizeof(uint8_t) * bytesToRead);
+        uint8_t bytesRead = (uint8_t) Serial.readBytes((char*)buffer, bytesRead);
```
ginywiny, line 14: Not sure if it is necessary to have an args length, since when each message is received (depending on the commandID type) there will be a fixed number of args for that specific command type (ex. fan_on uses only 1 arg, therefore reading the message will only read from 1 arg be default, and not require argsLength to say how many args to read from). 

Unless I am misunderstanding something.

MartensCedric: The purpose of this PR is just to create a basic architecture for you guys to work with. I just added these lines for you to know where to read/write commands. The details of how the commands are set should be tackled in the issue after #498

robot/rover/comms/src/Serial.cpp
MartensCedric: keep in mind that this needs to work with the PDS which doesn't use a teensy.

robot/rover/comms/src/CommandCenter.cpp
ginywiny: @MartensCedric Gotcha. I don't have anything else to add.

ginywiny (approved):
After receiving clarification, this looks like it can be pushed for further development.

VimFreeman (approved):
:+1:

PR #517, opened 2021-02-14
Pds2 comms 505

MewtR:
# Assignee Section

## Description
This PR doesn't solve anything in particular, but it shows how the new PDS functions can be written to work with the internal comms system.

* CMakelists.txt that is used to compile the new PDS and link it against the internal comms library.
* Example functions to disable/enable all motors, enable/disable a single motor or fan, ping/pong.
* Most example functions were simply copied from the first PDS and can/should be changed. They are basically stubs.
* Still lots of `Serial.print()`s for testing purposes.

### Steps for Testing
Note: You need an ATmega328p to run the tests. I used and Arduino Nano.

Other Note: To reduce noise, comment  out the code that reads the temperature in `PDS2.ino`, line 34-46.

1. Use the CMakeLists.txt provided in `robot/rover/PDS/PDS2` to compile the new pds.
2. Upload it to your device.
3. Make sure you have a way to view your devices output. For example `tail -f` technique from the sending bytes section.
4. Send raw bytes corresponding to a command. Refer to [here](https://docs.google.com/spreadsheets/d/1bE3h0ZCqPAUhW6Gn6G0fKEoOPdopGTZnmmWK1VuVurI/edit#gid=1682378273) for available commands
5. You should printed statements corresponding to the command you issued.

#### Additional info 
copied from [here](https://github.com/space-concordia-robotics/robotics-prototype/pull/516)

It's not trivial to use the Arduino IDE for the testing steps above so I had to do everything from the commandline. Here's how:

###### Compiling

Use the provided `CMakelists.txt`. These generate elf files

###### Uploading
The arduino ide will only upload a binary it compiled itself so we need to manually upload with `avrdude`. My command looked like the following `avrdude -C/home/bionic/arduino-1.8.12/hardware/tools/avr/etc/avrdude.conf -v -patmega328p -carduino -P/dev/ttyUSB0 -b57600 -D -Uflash:w:bin/PDS2.elf:e`. Just remember to change the device from `/dev/ttyUSB0` to whatever your device is and the path to `avrdude.conf` to the path on your system. Also put the path to your elf file in the last part : `-Uflash:w:<path/to/elf.elf>:e`

##### Sending bytes
The Arduino IDE's serial monitor only allows you to send ascii as far as I can tell. So I used python to send in raw bytes.
Basically in one terminal you should have `tail -f <device>` so for example `tail -f /dev/ttyUSB0`.
In another terminal you can run something like: `python3 -c "import sys; sys.stdout.buffer.write(b'\x36\x00\x02\x00\x02\x06\x64\x00')" > /dev/ttyUSB0` and you should see any results in the first terminal. If you don't see anything despite having some prints. Try sending data via the Arduino IDE's serial monitor first and then retry this method.

#### Example command:
`python3 -c "import sys; sys.stdout.buffer.write(b'\x35\x00\x02\x00\x02\x03\x01\x00')" > /dev/ttyUSB0` <- In this command:
- Command ID: 53 (single motor)
- Device sending: 0
- Device receiving: 2
- Argument length (2 bytes): 2
- Arguments: motor pin: 3, state: 1
- Stop byte: 0

closes nothing

# Reviewer Section


MartensCedric (approved):
There will need some changes to implement send commands in the future. Merging this into comms master because its good base code

PR #535, opened 2021-05-15
Flow Control 

VimFreeman:
# Assignee Section

## Description
Implemented flow control.
Will need to resolve conflicts bc of 42069

# Reviewer Section


requirements-rover.txt
```diff
@@ -28,3 +28,4 @@ Twisted==20.3.0
 pytest==3.6.3
 pylint==2.5.3
 yapf==0.30.0
+Jetson.GPIO
```
MartensCedric, line 31: no version?

MartensCedric: you can just pip freeze to find it I think

robot/rospackages/src/mcu_control/scripts/CommsNode.py
```diff
@@ -117,40 +113,44 @@ def send_command(command_name, args, deviceToSendTo):
...
+            for queue in command_queues:
+                if len(queue) > 0:
+                    cmd = queue.pop()
+                    send_command(cmd)
+
+            # Prompts each teensy to send something
+            for teensy in teensy_pins:
+                GPIO.output(teensy, 1)
+                receive_message()
+                GPIO.output(teensy, 0)
```
MartensCedric, line 148: receive_message calls a while(True): loop, are you sure you want to do that lol?

MartensCedric (approved):
Nice, remember what I told you about the Jetson making sure only one teensy talks on the bus at the time? The code is here @OfficialOwlElder

MartensCedric:
didn't test, this might not work at all, we might want to test the current comms-master-0 first, or maybe not, your judgement, we're going fast :sunglasses:

VimFreeman:
> didn't test, this might not work at all, we might want to test the current comms-master-0 first, or maybe not, your judgement, we're going fast :sunglasses: 

Yes totally should wait before merging, I realize it's missing some stuff that elec had to be done for

requirements-rover.txt
VimFreeman: I would rather keep the version to the latest. It's not like some day they will include breaking changes.

robot/rospackages/src/mcu_control/scripts/CommsNode.py
VimFreeman: No, not sure. I'll set a timeout.

VimFreeman: ok made it wait for half a second, hopefully that is plenty. I'm not sure if there is a better way. 
Also, maybe we can just set a global timeout when starting serial?

requirements-rover.txt
VimFreeman: (and even if they do, I would rather spend time staying up to date than be stuck on an obsolete version at some point)

robot/rover/internal_comms/src/CommandCenter.cpp
```diff
@@ -131,14 +138,16 @@ namespace internal_comms
...
-                //Serial.write(1);
-                //Serial.write(0);
-            //}
-        //}
+            else {
+                Serial.write(1);
+                Serial.write(0);
+                Serial.write(0x0A);
+            }
+        }
```
VimFreeman, line 146: This is probably gonna break, but it would be good to have an "empty" command, so that we don't timeout the while loop every time there is nothing to send

robot/rospackages/src/mcu_control/scripts/CommsNode.py
```diff
@@ -30,12 +33,14 @@ def get_handler(commandId, selectedDevice):
...
+# Pin definitions WHICH ARE WRONG
 ARM_PIN = 11
-WHEEL_PIN = 13
+ROVER_PIN = 13
 SCIENCE_PIN = 15
+teensy_pins = [ARM_PIN, ROVER_PIN, SCIENCE_PIN]
 
-teensy_pins = [ARM_PIN, WHEEL_PIN, SCIENCE_PIN]
+GPIO.setmode(GPIO.BOARD)
+GPIO.setup(teensy_pins, GPIO.OUT)
```
MartensCedric, line 43: please set all the pins to low at the start

MartensCedric (changes requested):

PR #559, opened 2022-05-18
Comms arm 540: New arm code

vexandmore:
# Assignee Section

## Description
Complete rewrite of the old arm code. Currently, only command that is implemented is set_motor_speeds.
It takes 6 floats, and works as follows:
- The first four are PWM values, [-255, 255] for the DC motors. The last two are for the servos, in RPM

Also, now sending messages back from a teensy to the OBC has been fixed and seems to be working. I have also added CommsNodeLocalTest.py , which is a version of commsnode that uses USB. You will need to change the serial port on it for it to work. It is using this that I have tested messages sending back to the obc.

The arm now sends back an message with ID 1 as a pong, and a message with ID 0 as a debug string when it encounters an invalid command.

### Steps for Testing

1. Compile and upload the arm code (see https://github.com/space-concordia-robotics/robotics-prototype/wiki/Building-the-C---code) then run upload Arm
2. Then, send a set motor speed command over serial to the Arm teensy. For instance, if the attached sketch [simulate-sending-motors.zip](https://github.com/space-concordia-robotics/robotics-prototype/files/8712657/simulate-sending-motors.zip) - is run in a teensy, which has Serial 1 connected to Serial 1 of the primary teensy, and grounds are connected - it will cause the first four motors, on pins 9, 5, 10, and 6, to pulse in sequence.

For the OBC things:
1. Compile and upload the Arm code, but in ArmDebug.h, uncomment `#define DEBUG`. 
2. In CommsNodeLocalTest, change line 225 with the serial port of your Teensy - change it from `/dev/ttyACM0` to the one you see in Arduino IDE.
3. Run ros master with `roscore`
4. Run this node with `rosrun mcu_control CommsNodeLocalTest.py`
5. In another terminal, run `rostopic echo /arm_feedback` 
6. In another terminal, run `rostopic pub /arm_command std_msgs/String "data: 'ping'"` you should see the LED on the teensy blink, and a pong response in arm_feedback
7. Run `rostopic pub /arm_command std_msgs/String "data: 'debug_test'"` You should see `data: "This is a debug string"` in arm_feedback
8. Run `rostopic pub /arm_command std_msgs/String "data: 'set_motor_speeds 255 0 0 0 0 0'" -r 11` if you plug a light into pin 9, you should see it light up.

# Reviewer Section


robot/rospackages/src/mcu_control/scripts/CommsNode.py
```diff
@@ -108,7 +108,7 @@ def receive_message():
                 print("No command with ID ", commandID, " was found")
                 ser.read_until() # 0A
 
-            argsLen = ser.read(2)
+            argsLen = ser.read()
```
steamypotato, line 111: Receive 1 byte here but CommandCenter.cpp sends two bytes (MSB and LSB) in the sendMessage() function. Could there be an issue here ?

vexandmore: To my knowledge, the args length changed from 2 bytes to 1 byte. So, sendMessage would be at fault here.

robot/rover/Arm/Arm.ino
```diff
@@ -1,894 +1,107 @@
...
-    motor4.pidController.setGainConstants(10.0, 0.0, 0.0);
-
-    // set motor shaft angle tolerances
-    motor1.pidController.setJointAngleTolerance(0.1);//2.0 * motor1.gearRatioReciprocal); // randomly chosen for dc
-    motor2.pidController.setJointAngleTolerance(0.1);//2.0 * motor2.gearRatioReciprocal);
-    motor3.pidController.setJointAngleTolerance(0.1);//2.0 * motor3.gearRatioReciprocal);
-    motor4.pidController.setJointAngleTolerance(0.1);//2.0 * motor4.gearRatioReciprocal);
-    motor5.pidController.setJointAngleTolerance(0.1);//2.0 * motor5.gearRatioReciprocal); // randomly chosen for servo
-    motor6.pidController.setJointAngleTolerance(0.1);//2.0 * motor6.gearRatioReciprocal);
+void invalidCommand(const uint8_t cmdID, const uint8_t* rawArgs,
```
VimFreeman, line 50: This is nice, but it will trigger for every single command that is for the wheels since the arm teensy will jump to the the default statement in the switch case for the command ids used by the wheels.

robot/rover/Arm/Arm.ino
```diff
@@ -1,894 +1,107 @@
...
-
-    // set motor joint angle limits
-    motor1.setAngleLimits(M1_MIN_HARD_ANGLE, M1_MAX_HARD_ANGLE, M1_MIN_SOFT_ANGLE, M1_MAX_SOFT_ANGLE);
-    motor2.setAngleLimits(M2_MIN_HARD_ANGLE, M2_MAX_HARD_ANGLE, M2_MIN_SOFT_ANGLE, M2_MAX_SOFT_ANGLE);
-    motor3.setAngleLimits(M3_MIN_HARD_ANGLE, M3_MAX_HARD_ANGLE, M3_MIN_SOFT_ANGLE, M3_MAX_SOFT_ANGLE);
-    motor4.setAngleLimits(M4_MIN_HARD_ANGLE, M4_MAX_HARD_ANGLE, M4_MIN_SOFT_ANGLE, M4_MAX_SOFT_ANGLE);
-    // motor5.setAngleLimits(M5_MIN_HARD_ANGLE, M5_MAX_HARD_ANGLE, M5_MIN_SOFT_ANGLE, M5_MAX_SOFT_ANGLE); // this joint should be able to spin freely
-    motor6.setAngleLimits(M6_MIN_HARD_ANGLE, M6_MAX_HARD_ANGLE, M6_MIN_SOFT_ANGLE, M6_MAX_SOFT_ANGLE);
+void pong() {
+  digitalWrite(LED_BUILTIN, !digitalRead(LED_BUILTIN));
```
VimFreeman, line 66: This only toggles the LED, so it will stay on/off after the pong. This is undesirable, so better to blink it or not have visual feedback. Same for invalidCommand()

robot/rover/Arm/src/commands/ArmCommandCenter.cpp
```diff
@@ -1,131 +1,77 @@
...
+      } else {
+        invalidCommand(cmdID, rawArgs, rawArgsLength);
+        break;
+      }
+    }
+    case COMMAND_DEBUG_TEST:
+      debug_test();
+      break;
+    default: {
+      invalidCommand(cmdID, rawArgs, rawArgsLength);
```
VimFreeman, line 74: This will trigger for all wheel commands. Maybe it can be wrapped in `#ifdef DEBUG`?

robot/rover/internal_comms/include/CommandCenter.h
```diff
@@ -5,9 +5,10 @@
 // And mostly written by Tim
 //
 
-#include "Arduino.h"
 #include <etl/queue.h>
 
+#include "Arduino.h"
```
VimFreeman, line 10: Please avoid formatting files automatically.

robot/rover/internal_comms/src/CommandCenter.cpp
```diff
@@ -4,167 +4,151 @@
...
+  unsigned long start = millis();
+  while (!Serial.available()) {
+    unsigned long current = millis() - start;
+    if (current > 50) break;
+  }
+  return Serial.read();
+}
 
+// No where else uses rxPin and txPin, BUT can't get rid of them without
+// breaking other code that is out of scope for me (Marc)
```
VimFreeman, line 86: It was designed like this for the PDS mcu but it might not be necessary anymore with PDS2. We will leave it since it is not causing any problems.

VimFreeman (changes requested):
This is very good work Marc. I can see that you put in a lot of time and effort into it.
There are a few concerns that I mention in the review and there are a couple of others that I will bring up saturday but I haven't mentioned bc it will be easier to discuss in person. :+1:

robot/rover/Arm/include/PinSetup.h
```diff
@@ -50,20 +50,7 @@
...
-#define M1_LIMIT_SW_CW        22 // LS 2
-#define M1_LIMIT_SW_CW_PORT   GPIOC_PDIR
-#define M1_LIMIT_SW_CW_SHIFT  CORE_PIN22_BIT
-
-#define M1_ENCODER_RESOLUTION 48
-#define M1_GEAR_RATIO         1.4*99.508*40.0 //TODO fix this based on gear reduction inside and outside motor
+
 #define M1_MIN_HARD_ANGLE     -175.0 // angle at which the join presses the limit switch
 #define M1_MAX_HARD_ANGLE     175.0
 #define M1_MIN_SOFT_ANGLE     -170.0 // hard angle with safety margin to avoid hitting limit switch after homing
```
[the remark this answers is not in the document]
vexandmore: these are the angles that were from the old arm code, I just left them in and deleted other stuff in PinSetup.h

robot/rover/Arm/src/commands/ArmCommandCenter.cpp
vexandmore: How is this undesirable?

